PR #59, opened 2018-09-14
Creates functions useful in URL building, and checks for presence of fragment identifier in URLs

### LINKED ISSUE #42: Error with redirect_uri parameters
The first url does not work, but the second does. The difference is the ordering of url parameters.

**Does not work**: https://accounts.google.com/o/oauth2/v2/auth?client_id=985459119454-r05c58fgdkk9o6sq7oupo3e9b44hb8q4.apps.googleusercontent.com&redirect_uri=https://localhost:8080/#/register/&scope=email&response_type=code

**Works**: https://accounts.google.com/o/oauth2/v2/auth?response_type=code&scope=profile%20email&client_id=985459119454-r05c58fgdkk9o6sq7oupo3e9b44hb8q4.apps.googleusercontent.com&redirect_uri=http://localhost:8080/#/register

The issue is in the parsing of the `redirect_uri` parameter when there is a `#` in the url. We should definitely put the `redirect_uri` parameter last in the url and we should also use a url building library instead of string concatenation.

https://github.com/HackIllinois/api/blob/5ef2301fc16447a435c84176a6105e4acd532aae/services/auth/service/oauth_service.go#L18

pattyjogal:
Fixes #42 

Turns out that if you have the `#` character **anywhere** in the URL query string, anything after it is not parsed, hence why we had to have it at the end all the time. I made a function that builds URLs and also checks to make sure `#` doesn't appear anywhere else but in the last query param. If it finds one, it passes out an error with the built URL string.


ASankaran:
As a general note. All PRs should be made into the `staging`, (or occasionally as feature branch) and never directly into `master`.

pattyjogal:
Right, my bad. I did base off of staging, I believe.

services/auth/service/oauth_service.go
```diff
@@ -146,3 +161,50 @@ func GetLastName(oauth_token string, provider string) (string, error) {
...
+	}
+
+	// Do some param validation to ensure that the # character doesn't appear in any values except the last
+	// This is because # nullifies the rest of the URL
+
+	valid := true
+	counter := 0
+	for _, val := range queryParams {
+		counter++
+		if strings.Contains(val, "#") && counter < len(queryParams) {
```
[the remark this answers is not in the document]
ASankaran: `https://tools.ietf.org/html/rfc6749#section-3.1.2`

We should reject URIs which include a fragment.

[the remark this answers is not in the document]
pattyjogal: Ok! That makes my job easier :smile:

services/auth/service/oauth_service.go
```diff
@@ -146,3 +161,50 @@ func GetLastName(oauth_token string, provider string) (string, error) {
...
+		Host:   host,
+		Path:   path,
+	}
+
+	// Do some param validation to ensure that the # character doesn't appear in any values except the last
+	// This is because # nullifies the rest of the URL
+
+	valid := true
+	counter := 0
+	for _, val := range queryParams {
```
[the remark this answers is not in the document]
pattyjogal: Ah, I think the fact that queryParams is a map makes it return the key instead! I couldn't find a better way (like Python's enumerate), but if you know any, let me know!

services/auth/tests/url_builder_test.go
```diff
@@ -0,0 +1,109 @@
...
+   Test that a simple URL with no query params can be generated
+*/
+func TestConstructSimpleURL(t *testing.T) {
+	result, err := service.ConstructSafeURL(URL_SCHEME, URL_HOST, URL_PATH, nil)
+
+	if result != CONSTRUCTED_BASIC_URL {
+		t.Errorf("URL not correctly constructed. Expected \"%v\", got \"%v\"", CONSTRUCTED_BASIC_URL, result)
+	}
+
+	if err != nil {
```
ASankaran: In general you should be checking `err != nil` first before checking `result`. This is true for all functions which return some values then an `err`.

services/auth/service/oauth_service.go
```diff
@@ -146,3 +163,49 @@ func GetLastName(oauth_token string, provider string) (string, error) {
...
+		Path:   path,
+	}
+
+	// Per the OAuth 2.0 RFC 6749, we need to disallow the `#` fragment character in the URL
+	valid := true
+	if queryParams != nil {
+		for _, val := range queryParams {
+			if strings.Contains(val, "#") {
+				valid = false
+				break
```
ASankaran: Why not just early return here instead of having a bool for control flow?

services/auth/tests/url_builder_test.go
```diff
@@ -0,0 +1,109 @@
...
+const DIFFERENT_SCHEME = "ftp"
+const HOST_WITH_PORT = "hackillinois.com:9800"
+const CONSTRUCTED_COMPLEX_URL = "ftp://hackillinois.com:9800/register/now"
+
+var QUERY_PARAMS = map[string]string{
+	"major":           "CS",
+	"graduating_year": "2021",
+}
+
+const EXPECTED_QUERY_STRING = "?graduating_year=2021&major=CS"
```
ASankaran: Can you guarantee the ordering of the parameters in these expected query strings? As far as I know the functions which encode a `map[string]string` to a query string do not guarantee an ordering.

services/auth/tests/url_builder_test.go
```diff
@@ -0,0 +1,109 @@
...
+   can be generated.
+*/
+func TestConstructComplexURL(t *testing.T) {
+	result, err := service.ConstructSafeURL(DIFFERENT_SCHEME, HOST_WITH_PORT, URL_PATH, nil)
+
+	if result != CONSTRUCTED_COMPLEX_URL {
+		t.Errorf("URL not correctly constructed. Expected \"%v\", got \"%v\"", CONSTRUCTED_COMPLEX_URL, result)
+	}
+
+	if err != nil {
```
ASankaran: Same idea with checking `err != nil` first.

services/auth/tests/url_builder_test.go
```diff
@@ -0,0 +1,109 @@
...
+func TestQueryStringBuilder(t *testing.T) {
+	url := url.URL{
+		Scheme: URL_SCHEME,
+		Host:   URL_HOST,
+		Path:   URL_PATH,
+	}
+
+	service.ConstructURLQuery(&url, QUERY_PARAMS)
+
+	queryString := "?" + strings.Split(url.String(), "?")[1]
```
ASankaran: If the string which is returned by `url.String()` doesn't have a `?` this will panic. While I agree that the string should always have a `?` tests should not assume anything.

services/auth/tests/url_builder_test.go
```diff
@@ -0,0 +1,109 @@
...
+*/
+func TestConstructFullURL(t *testing.T) {
+	result, err := service.ConstructSafeURL(URL_SCHEME, URL_HOST, URL_PATH, QUERY_PARAMS)
+
+	expectedURL := CONSTRUCTED_BASIC_URL + EXPECTED_QUERY_STRING
+	if result != expectedURL {
+		t.Errorf("URL not correctly constructed. Expected \"%v\", got \"%v\"", expectedURL, result)
+	}
+
+	if err != nil {
```
ASankaran: Same thing with checking `err != nil` first.

services/auth/service/oauth_service.go
pattyjogal: Sorry, that was an artefact of my earlier loop idea, and I forgot to remove it!

services/auth/tests/url_builder_test.go
pattyjogal: I decided to fix this by testing the complete URL with just one query param, to guarantee the output. I also fixed the multi-param test by checking the url.ParseQuery() result against a known url.Values containing the params.

pattyjogal:
@ASankaran I went through your review and fixed everything! Let me know if it needs anything else!

PR #67, opened 2018-09-27
[WIP] Add apirequest package to common; Fix socket leaks

### LINKED ISSUE #54: Fix socket leaks
Currently there are many `http.Get`, `http.Post`, etc calls where we never close the request body causing the the socket for the connection to remain open. Eventually services run out of available sockets and start rejecting connections.

The best long term fix is to use a higher level requests library that manages these issues for us or to add wrapper functions to our common library which defer closing the socket.

ASankaran:
[WIP] Fixes #54

Currently this PR just adds the `apirequest` package to `common`. This package should be used for all inter-service communication. The goal here is to eliminate the issue of leaving a socket open.

- [x] Adds the `apirequest` package to `common`
- [x] Use `apirequest` package to fix socket leaks

### DIFF
--- a/common/apirequest/apirequest.go
+++ b/common/apirequest/apirequest.go
@@ -11,6 +11,8 @@ const (
 	Timeout int64 = 10
 )
 
+var Identity = ""
+
 /*
 	Executes an API request and populates the data with the response
 */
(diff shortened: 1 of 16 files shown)


common/apirequest/apirequest.go
```diff
@@ -0,0 +1,81 @@
...
+	Executes an API DELETE request and populates the data with the response
+*/
+func Delete(url string, data *interface{}) (int, error) {
+	return doRequest("DELETE", url, nil, data)
+}
+
+/*
+	Builds a request, executes it, and then decodes the response into data
+*/
+func doRequest(method string, url string, payload *interface{}, data *interface{}) (int, error) {
```
PradyumnaShome: Perhaps change `data` to `response`, to clarify what it represents?

PradyumnaShome:
Are the headers required? I assume that for inter-service API requests, both ends will already know the `userId`.

ASankaran:
Setting `Content-Type` is good practice and we do use `HackIllinois-Identity` when setting the user's initial decisions to pending.

ASankaran:
Since this PR changes all inter-service communication I'd like to get a couple people to look it over and to try out endpoints just to make sure nothing panics.

PR #68, opened 2018-09-27
switch format to RAW for UpdateCurrentUploadInfo

### LINKED ISSUE #47: Format for resume upload is incorrect
This endpoint should be using arbor's `RAW` format instead of the default `JSON` format we use everywhere else.

https://github.com/HackIllinois/api/blob/5ef2301fc16447a435c84176a6105e4acd532aae/gateway/services/upload.go#L41

Bo887:
Fixes #47

Switches the format from UpdateCurrentUploadInfo to RAW from JSON.

### DIFF
--- a/gateway/services/upload.go
+++ b/gateway/services/upload.go
@@ -11,6 +11,7 @@ import (
 var UploadURL = config.UPLOAD_SERVICE
 
 const UploadFormat string = "JSON"
+const RawFormat string = "RAW"
 
 var UploadRoutes = arbor.RouteCollection{
 	arbor.Route{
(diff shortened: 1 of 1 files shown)


gateway/services/upload.go
```diff
@@ -11,6 +11,7 @@ import (
 var UploadURL = config.UPLOAD_SERVICE
 
 const UploadFormat string = "JSON"
+const RawFormat string = "RAW"
```
PradyumnaShome: Change RawFormat to UploadFormat, since that is the format used to actually upload files via a PUT request, and change UploadFormat to InfoFormat, since that is used only to query information. Update the usage as well.

Issue #58, opened 2018-09-14
Allow people to check-in to events only when they are active


PradyumnaShome:
Current;y we do not check the event start and end times when checking attendees into events. Attendees should only be able to check-in to an event only if is currently occurring. 

- [ ] Verify, using the current time during the request, that a user is allowed to be marked as checked-in.

itsmesatwik:
I believe the event microservice uses the "MarkUserAsAttendingEvent" for checking in a particular event, we could directly include a function "HasEventStarted" for getting the required info for the the condition in event_service.go for checking in only when event is going on.
@PradyumnaShome

itsmesatwik:
Additionally we could possibly include a check-in valid function for checking in only till the first n-minutes after which we don't check in, to prevent unnecessary walk ins that disturb the talks.

PR #69, opened 2018-09-27
Mark users as attending only once event has started

itsmesatwik:
[Issue](https://github.com/HackIllinois/api/issues/58)
Added a function for validating the start of an event `CanCheckInEvent(string eventName)` and used that as a precondition in `MarkUserAsAttendingEvent` for only marking attending if event has started.
screen shot 2018-09-26 at 10 09 38 pm

### DIFF
--- a/services/event/service/event_service.go
+++ b/services/event/service/event_service.go
@@ -44,7 +44,7 @@ func CanCheckInEvent(name string) (bool, error) {
 	startTime := event.StartTime
 	t := time.Now().Unix()
 
-	if int(t)  < (int(startTime) - startCheckInTime) {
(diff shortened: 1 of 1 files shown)


itsmesatwik:
I understand that I'm getting some type incompatibility, still waiting on a response from @ASankaran for clarifying how we're using time, once I get that, fixes are easy.

YashoSharma:
Give this a readable name

PradyumnaShome:
FYI you can run `go test` locally.

YashoSharma:
Branch is stale and not being updated. Reopen if progress is made on this.

PR #71, opened 2018-09-30
Write resume uploads to /tmp/ when running locally

### LINKED ISSUE #6: Write uploads to /tmp when not running in production mode
- [x] Use the environment variable `IS_PRODUCTION` to determine if running in production
- [x] If in production `PUT` requests should upload files to the specified s3 bucket and `GET` requests should return a presigned url authorizing the downloading of the file from the s3 bucket
- [x] If not in production `PUT` requests should write the file to disk in a location on /tmp and `GET` requests should return the local path of the file

The logic for determining if running in production should be done at the service level. I recommend having separate functions for the local file handling and having that code path get executed instead of the current s3 code path when not running in production.

benpankow:
Resolves #6.

If the `IS_PRODUCTION` environment variable is not set to `true`, resumes uploaded through the PUT endpoint will be written to `/tmp/uploads/{id}.pdf`. The resume-related GET endpoint will return a local path as well.

Modifies the Travis configuration to force usage of the dev Arbor branch (mimics [similar change on the R|P API](https://github.com/ReflectionsProjections/api/pull/4)). Without this, we can't utilize the RAW format.

Documentation is left unchanged for now since the existing resume docs don't mention s3.

### DIFF
--- a/services/upload/service/upload_service.go
+++ b/services/upload/service/upload_service.go
@@ -9,9 +9,10 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
-	"net/http"
-	"time"
 	"io/ioutil"
+	"net/http"
+	"os"
+	"time"
 )
 
 var sess *session.Session
@@ -46,7 +47,7 @@ func GetUserResumeLink(id string) (*models.UserResume, error) {
 			return nil, err
 		}
 	} else {
-		signed_url = "/tmp/" + id + ".pdf"
+		signed_url = "/tmp/uploads" + id + ".pdf"
 	}
 
 	resume := models.UserResume{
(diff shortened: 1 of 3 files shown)


services/upload/service/upload_service.go
```diff
@@ -58,12 +68,18 @@ func UpdateUserResume(id string, file_buffer []byte) error {
...
+	if config.IS_PRODUCTION {
+		_, err = uploader.Upload(&s3manager.UploadInput{
+			Bucket:      aws.String(config.S3_BUCKET),
+			Key:         aws.String("resumes/" + id + ".pdf"),
+			Body:        bytes.NewReader(file_buffer),
+			ContentType: &content_type,
+		})
+	} else {
+		os.Mkdir("/tmp/uploads", os.ModePerm)
+		err = ioutil.WriteFile("/tmp/uploads/"+id+".pdf", file_buffer, 0644)
```
ASankaran: We have spaces before the `+` everywhere else, so we should have it here as well.

ASankaran: Also pull the `0644` out to a well named constant since there's a chance other developers might not know unix filesystem permissions well.

PR #72, opened 2018-10-01
Enumerated Roles

### LINKED ISSUE #37: Enumerate roles
Create enum for roles and use that instead of string representations.

ethan-lord:
Fixes #37


ethan-lord:
I modified the existing tests that used roles to test with the new enums, are there other tests I should add?

ASankaran:
So in terms of testing the next steps would be to add the role constants to the gateway service. Then you can unit test each endpoint to check it has the correct access.

YashoSharma:
Imo I’d define the roles at the gateway level then integrate test with endpoints added at runtime @ASankaran that should be doable right? We want to check that the roles are handled correctly on the gateway not that the services are setting the right roles. This avoids creating linkage between those components. Service level integration tests would handle testing the role on each of their own endpoints.

ASankaran:
Individual services don't have a notion of permissions. The service assumes that any request it receives is authorized. If the request was not authorized when the gatway would have rejected it before it reached the service.

It looks like @ethan-lord has added the enumeration of roles for when we give a user additional permissions.

Issue #37 essentially has two parts to it I think. The first being to remove the usage of strings everywhere and enumerate the roles as a set of constants which has been done here. The second is to add this enumeration to the gateway and test that the authorization middleware correctly accepts / rejects requests based on the user's roles and the endpoint's allowed roles.

Right now it looks like part 1 of the issue is completed. For part 2: Yes, I think we should have additional testing at the gateway level to ensure requests are correctly accepted / rejected.

ASankaran:
Closing this for now due to age. Feel free it reopen in the future once it is updated.

PR #74, opened 2018-10-06
Migrated QR codes to user service, from checkin service.

PradyumnaShome:
Moved controllers
Moved model
Moved services
Moved tests
Moved gateway routes
Moved documentation


ASankaran:
`services/auth/docs/Authorization.md` still has bad line endings

gateway/services/user.go
```diff
@@ -39,10 +51,22 @@ var UserRoutes = arbor.RouteCollection{
 	},
 }
 
+func GetCurrentUserInfo(w http.ResponseWriter, r *http.Request) {
```
ASankaran: Is this function used anywhere?

PR #76, opened 2018-10-13
Add apirequest package to common; Fix socket leaks

### LINKED ISSUE #54: Fix socket leaks

ASankaran:
Fixes #54

Currently this PR just adds the `apirequest` package to `common`. This package should be used for all inter-service communication. The goal here is to eliminate the issue of leaving a socket open.

- [x] Adds the `apirequest` package to `common`
- [x] Use `apirequest` package to fix socket leaks

### DIFF
--- a/common/apirequest/apirequest.go
+++ b/common/apirequest/apirequest.go
@@ -11,6 +11,8 @@ const (
 	Timeout int64 = 10
 )
 
+var Identity = ""
+
 /*
 	Executes an API request and populates the data with the response
 */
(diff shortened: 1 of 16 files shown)


YashoSharma:
Fix merge conflicts, also it would be nice that the base level requests function has a test that asserts the socket is closed (not sure how feasible this is with go's testing mechanics).

Issue #62, opened 2018-09-19
Refactor InitMongoDatabase to Connect instead


shreychowdhary:
Refactor InitMongoDatabase to Connect in database.go, so it is a method of the database struct.

Issue #36, opened 2018-08-22
Refine error handling


YashoSharma:
API integration for consumers would be made easier if error handling was unified across the codebase and errors were presented with a hierarchical order e.g. `ApiError` indicates there is an unhanded error, `UnprocessableRequestError` suggests that the request a user had made is not valid, etc..

My proposed solution for this is that:
- [ ] Custom error types are created for the api
- [ ] Current calls to `error.New` and usage of library errors are replaced with our own error types
- [ ] All error types have a `type` field to allow for easy checking of what error was returned by the API
- [ ] All error types have a `source` field

PR #80, opened 2018-10-15
Refactor commons/database

### LINKED ISSUE #35: Make common/database generic
This package was originally designed to be generic, but over time mongo specific types became a part of the interface. The `Database` interface should be refactored to not return `*mgo.ChangeInfo`. Instead we should return our own struct that is similar to `*mgo.ChangeInfo`. Structs which implement the interface will need to populate this custom ChangeInfo like struct. Additionally while the `Database` interface does return a generic `error` type, users of the interface currently import `mgo.ErrNotFound` to check if the returned error was a not found error. We should have our own Not Found error that can be returned and checked against instead. We may also need to provide other common error types as well. We will also need to find a way to abstract away the need to pass in `bson.M` structs into the query. We can use `map[string]interface{}` instead.

- [x] Remove mongo specific return types for the `Database` interface
- [x] Create a set of common error types to return instead of database specific errors
- [x] Abstract away the use of query types much as `bson.M`

ASankaran:
Fixes #35, #62 

This PR refactors the `common/database` to abstract away the particular database we choose to use. This allows us to easily change the database implement without changing any of the code in the micro services.

- [x] Change return types on `Database` interface methods to remove `mgo` specific types
- [x] Change errors returned to use `common/database` errors instead of `mgo` specific errors (This is the first step for issue #36)
- [x] Add a `QuerySelector` type to remove usage of `bson.M`

All unit tests are passing and in theory all of our database operations have unit tests, however additional manual testing it required before merging.


ASankaran:
Merging this now since we the number of open PRs which will require changes of this is growing and I haven't had time to write the db interface tests yet. Those tests will be added in a later PR.

PR #81, opened 2018-10-15
Add endpoint to get all checked in users

### LINKED ISSUE #66: Add GetAllCheckedInUsers Endpoint
Right now we have no way to retrieve a list of all checkedin users. That would be a very useful feature to have.

- [ ] Add endpoint to retrieve a list of all checkedin users

benpankow:
Resolves #66.

- [x] Adds an endpoint `/checkin/list/` that returns a list of all checked in user IDs. Requires Admin permissions.
- [x] Adds information about the endpoint to docs
- [x] Adds unit tests


ASankaran:
Updated PR to use new database interface.

Issue #61, opened 2018-09-17
Add users to a registered mailing list upon registration completion


ASankaran:
Currently we only add users to mailing lists once we finalize a decision for them in the decision service. It would be helpful to also have a mailing list of all registered users.

- [x] Add user to `registered` mailing list upon registration completion

PR #83, opened 2018-10-18
Add users to a registered user mailing list, after registration.

PradyumnaShome:
Addresses #61.

Users are added to a mailing list once they are registered.


PradyumnaShome:
I forgot to check if a mailing list exists. I'll rework that and get back.

services/registration/service/mail_service.go
```diff
@@ -30,3 +30,37 @@ func SendUserMail(id string, template string) error {
...
+	}
+
+	update_request_body := bytes.Buffer{}
+	json.NewEncoder(&update_request_body).Encode(&mail_list)
+
+	content_type := "application/json"
+
+	_, err := http.Post(add_to_mail_list_url, content_type, &update_request_body)
+
+	if err != nil {
```
ASankaran: If the mailing list doesn't exist err will still be nil since the request happened without error. You'll need to check the code on the response to see if the user was successfully added to the list.

services/registration/service/mail_service.go
```diff
@@ -25,3 +27,35 @@ func SendUserMail(id string, template string) error {
...
+		ID:      mail_list_id,
+		UserIDs: []string{user_id},
+	}
+
+	update_request_body := bytes.Buffer{}
+	json.NewEncoder(&update_request_body).Encode(&mail_list)
+
+	status, err := apirequest.Post(add_to_mail_list_url, &update_request_body, nil)
+
+	if err != nil || status != http.StatusOK {
```
ASankaran: If `err != nil` that indicates a different error than `status != http.StatusOK` and need to be handed differently.

You can add
```
if err != nil {
    return err
}
```

above that check for status.

services/registration/service/mail_service.go
```diff
@@ -25,3 +27,35 @@ func SendUserMail(id string, template string) error {
...
+		// The mailing list didn't exist
+		create_mail_list_url := fmt.Sprintf("%s/mail/list/create/", config.MAIL_SERVICE)
+
+		create_request_body := bytes.Buffer{}
+		json.NewEncoder(&create_request_body).Encode(&mail_list)
+
+		_, err := apirequest.Post(create_mail_list_url, &create_request_body, nil)
```
ASankaran: You should check the status code on the response here.

YashoSharma:
Updates?

services/registration/service/mail_service.go
```diff
@@ -25,3 +28,47 @@ func SendUserMail(id string, template string) error {
...
+
+		if err != nil {
+			return err
+		}
+
+		if status != http.StatusOK {
+			return errors.New("Mailing list does not exist, furthermore, creation of mailing list failed.")
+		}
+
+		return err
```
ASankaran: `return nil`

Might as well be explicit here since err can only be nil.

services/registration/service/mail_service.go
```diff
@@ -25,3 +28,47 @@ func SendUserMail(id string, template string) error {
...
+	}
+
+	return err
```
ASankaran: `return nil`

PR #84, opened 2018-10-24
Modify role-setting endpoints

### LINKED ISSUE #28: Change endpoints for setting roles
Currently the auth service has endpoints for retrieving the roles of a user as a list and setting the list of roles the user has. The endpoint which sets the roles of a user to a given list should be removed. Instead two new endpoints should be added. One which takes a single role and adds it to the user and one which takes a single role and removes it from a user.

- [ ] Remove `PUT /auth/roles/`
- [ ] Add `PUT /auth/roles/add/`
- [ ] Add `PUT /auth/roles/remove/`

benpankow:
Resolves #28.

Removes the `PUT /auth/roles/` endpoint and replaces it with `PUT /auth/roles/add/` and `PUT /auth/roles/remove/`. Requires Admin permissions.

Adds relevant documentation and tests.

Request format:

```
{
	"id": "github6892396",
	"role": "User"
}
```

Response format:

```
{
	"id": "github6892396",
	"roles": [
		"User"
	]
}
```

`/roles/remove/` will throw a `400` response if the user does not have the specified role.

References to the old endpoint were updated.

In addition, two slice utility functions were created and added to `common/utils/slice_utils.go`, `ContainsString` and `RemoveString`, which check if a slice contains a string and remove a string from a given slice, respectively.

### DIFF
--- a/services/auth/docs/Authorization.md
+++ b/services/auth/docs/Authorization.md
@@ -1,8 +1,6 @@
-Authorization
-=============
+# Authorization
 
-GET /auth/PROVIDER/?redirect_uri=AUTHREDIRECTURI
-------------------------------------------------
+## GET /auth/PROVIDER/?redirect_uri=AUTHREDIRECTURI
 
 Redirects to the `PROVIDER`'s OAuth authorization page. Once the user accepts the OAuth authorization they will be redirected to the client's auth page with an OAuth code. This code should be sent to the API to be exchanged for an API JWT.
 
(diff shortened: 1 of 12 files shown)


services/auth/docs/Authorization.md
```diff
@@ -1,45 +1,45 @@
-Authorization
-=============
+# Authorization
```
ASankaran: What's the reasoning for changing the method of creating titles? By changing it only here this file becomes inconsistant with the rest of the API.

benpankow: My mistake - was trying out a save-on-format plugin for something else and didn't realize that it also formatted markdown files. Fixed.

PR #85, opened 2018-10-24
Make user an Attendee on check-in override

### LINKED ISSUE #11: Make user an Attendee on check-in override
If a user is given an override to check-in they should have the same permissions in the API as anyone else who checked in.

- [ ] Add the role `Attendee` to a user upon checkin override

benpankow:
Depends on #84, resolves #11.

Adds the "Attendee" role to a user if they at any point receive a check-in override (via `POST /checkin/` or `PUT /checkin/`) so that they have the same status as an RSVP-ed attendee.

Tested manually by hitting both relevant endpoints.

### DIFF
--- a/services/checkin/service/auth_service.go
+++ b/services/checkin/service/auth_service.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"github.com/HackIllinois/api/common/apirequest"
 	"github.com/HackIllinois/api/services/rsvp/config"
 	"github.com/HackIllinois/api/services/rsvp/models"
 	"net/http"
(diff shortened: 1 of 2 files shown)


services/checkin/service/auth_service.go
```diff
@@ -0,0 +1,55 @@
...
+		return errors.New("Auth service failed to update roles")
+	}
+
+	return nil
+}
+
+/*
+	Remove Attendee role from user with auth service
+*/
+func RemoveAttendeeRole(id string) error {
```
ASankaran: Why do we need this function? It doesn't seem to be used anywhere in this service.

benpankow: Fixed.

PR #86, opened 2018-10-25
High quality, easy to read, documentation

PradyumnaShome:
- [x] Define how we will be serving documentation to API consumers and how we will be generating this documentation
> MkDocs, which creates a static website in the `/documentation/site/` directory. This can be added to the CI / CD pipeline via an S3 bucket, such that merges into `master` are propagated to the live website.
- [x] Write the updated documentation. Dependencies include Python and other packages for which a Makefile has been added.
> Currently, using a collection of existing Markdown files written. More information can be included in a later pull request.
- [x] Update the build system to deploy documentation on merges into master
> Pending


ASankaran:
Since you're moving all of the docs to `documentation/docs/reference` you can also delete all of the `docs` folder from each service.

ASankaran:
Looks like there are some more conflicts on this PR. Can you resolve those?

ASankaran:
This looks good. One minor addition that would be nice to have is a `make documentation` target which can be run from the root of the repository that installs dependencies if needed and builds the static contents for the site. You can probably just add the `docs` target to our top level Makefile and call make through to `documentation/`.

ASankaran:
One more minor request. It looks like `documentation/docs/getting-started/developer-guide.md` is mostly the same as the readme file. Can you set something up so that changes in the README propagate to that file.

Makefile
```diff
@@ -49,3 +49,7 @@ setup: all
 	@echo 'Generating API admin token'
 	hackillinois-utility-accountgen
 	hackillinois-utility-tokengen
+
+.PHONY: docs
+docs:
+	cd $(BASE_DIR)/documentation && $(MAKE) -C build && cd -
```
ASankaran: Does this work? The purpose of the `-C` flag is so that we don't need to change directories to run make.

PradyumnaShome: I guess I'm not sure what $(MAKE) -C does. What does it do, and how does it know what target to hit / Makefile to run on?
```suggestion
	$(MAKE) -C build
```

ASankaran: https://www.gnu.org/software/make/manual/html_node/Options-Summary.html

You'll probably want something along the lines of:
`$(MAKE) -C $(BASE_DIR)/documentation build`

But test it and make sure it works.

YashoSharma:
Updates on this? This pull should also include the diff for the travis config

Makefile
PradyumnaShome: Yes this works.

PradyumnaShome:
Will attempt to add relevant targets to the `travis.yml`.

PradyumnaShome:
@ASankaran Maybe I can move the dependency fetching out of the `Makefile`?
Would it be a good idea to add the commands to the `install` stage?

documentation/Makefile
```diff
@@ -1,14 +1,19 @@
 SOURCES = mkdocs.yml $(wildcard */*.md) 
 DOCS_ADDENDUM = ../README.md
+REQUIREMENTS = requirements.txt
 
-.PHONY: serve build copy-docs-addendum
+.PHONY: install-deps serve build copy-docs-addendum
+
+# Install dependencies
+install-deps: $(REQUIREMENTS)
+	pip install -r $(REQUIREMENTS)
```
ASankaran: I think this should be `pip3`?

Issue #27, opened 2018-08-07
Define level of access for Staff members in the API


ASankaran:
Currently the Staff role is not actually used in the gateway for authenticating requests. There are a number of endpoints that are currently restricted to Admin users that can be opened up to Staff as well.

- [ ] Define endpoints that staff should have access to
- [ ] Update these endpoints in the gateway

PR #88, opened 2018-11-01
Add staff role

ASankaran:
Addresses #27 and #37 

- [x] Define and update routes which should be accessible to all staff

I have left all mail and event creation / deletion routes as `Admin` for now. We can consider adding a `Senior Staff` role in the future which has these permissions.


ASankaran:
Thoughts on the permissions for mail and event creation / deletion? @YashoSharma

YashoSharma:
Will look when I get home.

ASankaran:
#72 is getting stale, so I added role enumeration to the gateway in this PR.

PR #89, opened 2018-11-02
Use presigned s3 urls for uploads

### LINKED ISSUE #15: Use presigned urls for uploads
Right now we use presigned urls for downloading of files. We should also use them for uploading files. This would eliminate the need to move large files though our gateway and upload services greatly reducing load on the API.

- [x] Upload files using pre-signed URLs

ASankaran:
Fixes #15, (and makes #5 obsolete)

This PR changes the way we upload resumes. Originally API users would upload the file to our API and we would upload that to S3. There's isn't a reason we should need to relay the data. So now users can request a presigned url for resume uploads from the API and then upload the file directly to S3.

Potential additions to look into are enforcing a `Content-Type` and max `Content-Length` for uploads using the presigned url.

- [x] Upload resumes via presigned urls


ASankaran:
Looks the upload service never had tests, since there wasn't much at the service level. And anyways the tests take the non-production path which doesn't have anything useful to test in with this upload.

Add tests would be good, but I'm not sure exactly what to be testing here.

PR #93, opened 2018-11-15
Added checking for events being active while scanning in attendees.

PradyumnaShome:
Addresses #58.

Included tests.

Will document this feature once the new docs are merged in, and the build passes.

As part of this feature, I had to edit the start and end times of all the other events used in other tests, so that people could be checked in.


services/event/service/event_service.go
```diff
@@ -286,3 +298,22 @@ func MarkUserAsAttendingEvent(event_name string, user_id string) error {
...
+	event, err := GetEvent(event_name)
+
+	if err != nil {
+		return false, err
+	}
+
+	startTime := event.StartTime
+	t := time.Now().Unix()
+
+	return math.Abs((float64)(t-startTime)) <= CheckInTimeIntervalInSeconds, nil
```
ASankaran: We might have some long events (1+ hours). I'd say people should be able to checkin to the event up until the end of the event. We can probably also shorten up the pre-event checkin time. I'd say 15 minutes before an event till the end of the event is the time frame we want.

Issue #49, opened 2018-09-03
Services startup when critical environment variables are missing


YashoSharma:
- Services should not start up if a necessary environment variable is missing

PR #94, opened 2018-11-18
Config loader

ASankaran:
Addresses #49 

Currently when deploying to AWS we need to manually update all the environment variable associated with each given container. This is makes it easy to miss adding a variable to a deployment. Most of the time these environment variables are not secrets, so we should be able to automatically deploy them.

A solution to this is to store key-value pairs in a json files which can be loaded from an s3 uri, filepath, or https url. Once loaded environment variables will still take precedence so secrets can be loaded into the services.

Currently this PR adds the `ConfigLoader` with the following interface
```
func Load(configPath string) (*ConfigLoader, error)
func (loader *ConfigLoader) Get(key string) (string, error)
func (loader *ConfigLoader) ParseInto(key string, out interface{}) error
```

For now we will only make use of the `Load` and `Get` methods. The `ParseInto` method gives us the ability to store complex data in our config files, which will be useful when implementing the generic registration interface so that HackIllinois and R|P can be deployed off of the same repository.

In production we can load from a file in our s3 bucket and in development we can load from a local file. The config location will be loaded into containers as the only necessary environment variable (besides secrets).

This addresses #49, by terminating services when they do not have the required configuration variables set.

- [x] Implement `ConfigLoader`
- [x] Test `ConfigLoader`
- [x] Integrate into services


common/configloader/configloader.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"os"
+)
+
+var ErrNotSet = errors.New("The value for the given key was not set")
```
PradyumnaShome: `const`?

common/configloader/configloader.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return value, nil
+}
+
+/*
+	Parses the value the given key into the given interface{}
+	Environment variables will override configutation
+*/
+func (loader *ConfigLoader) ParseInto(key string, out interface{}) error {
```
PradyumnaShome: What is the purpose of this method?

common/tests/configloader_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	err = ioutil.WriteFile("/tmp/testconfig.json", cfg_json, 0644)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func Teardown(t *testing.T) {
```
PradyumnaShome: `TearDown`?

common/tests/configloader_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	"encoding/json"
+	"github.com/HackIllinois/api/common/configloader"
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func Setup(t *testing.T) {
```
PradyumnaShome: `SetUp`?

ASankaran: `Setup` is one word.

common/tests/configloader_test.go
ASankaran: `Teardown` is one word.

PradyumnaShome: This is not very important, so up to you. I think [JUnit](https://junit.org/junit4/faq.html#organize_3) uses different conventions, which is why I brought it up.

common/configloader/configloader.go
ASankaran: If we have json more complex than a key-value pair, then it is impossible to parse until we have a struct definition to parse into. So we when loading the json we parse the top level and and further nesting can be parsed with this method. For an example of it's usage you can see the associated test.

common/configloader/configloader.go
ASankaran: `errors.New(...)` isn't evaluated at compile time so it is not `const`.

ASankaran:
The purpose of this is not to replace `config.go`, but to make it much simpler to configure containers in production. The interface exported by a service's `config` package would be exactly the same, except we load via the `ConfigLoader` instead of directly from environment variables. This allows us to avoid breaking existing code to and we still only need to load configuration variables when the config package is imported.

We can probably test loading from http via a mock using `jarcoal/httpmock`. I am unfamiliar with mocking S3, so that might be more difficult.

common/configloader/configloader.go
```diff
@@ -0,0 +1,173 @@
...
+	uri, err := url.Parse(configPath)
+
+	if err != nil {
+		return nil, err
+	}
+
+	region, exists := os.LookupEnv("S3_REGION")
+
+	if !exists {
+		region = "us-east-1"
```
PradyumnaShome: What if the event using the API doesn't have an S3 bucket in North Virginia?

Isn't this something we want to eliminate using the `configloader` package?
I guess this a recursive issue (we need a config for the API, and a "config" to load the config).

ASankaran:
Loading config files from S3 is tested and working.

benpankow:
Looks good to me :+1:

Issue #53, opened 2018-09-06
Health Checks


YashoSharma:
- [x] Health checks would allow services to mitigate the effect of task definitions that suffer from unexpected bugs that render them unable to serve further requests

PR #123, opened 2018-12-31
Service health endpoints

ASankaran:
Addresses #53 

Adds a new endpoint to each service at `/{service_name}/internal/healthstats/` which reports information on average response time, http response status codes, etc.

This PR adds a new `apiserver` package to common which handles the common server startup code for each service. Requests also now have a request timeout of 10 seconds.

If the average response time get too high or we see an unusual number of non 200 return status codes, we can have the health check fail and restart the container.

### DIFF
--- a/common/apiserver/apiserver.go
+++ b/common/apiserver/apiserver.go
@@ -1,0 +1,14 @@
+package apiserver
+
+import (
+	"net/http"
+	"github.com/gorilla/mux"
+	"github.com/HackIllinois/api/common/middleware"
+)
+
+func StartServer(address string, router *mux.Router) error {
+	router.Use(middleware.ErrorMiddleware)
+	router.Use(middleware.ContentTypeMiddleware)
+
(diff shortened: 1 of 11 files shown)


ASankaran:
Sample response from a health stats endpoint:
```
{
    "pid": 23367,
    "hostname": "",
    "uptime": "1m22.12548s",
    "uptime_sec": 82.12548,
    "time": "2018-12-31 13:51:38.2300216 -0600 STD m=+82.157222001",
    "unixtime": 1546285898,
    "status_code_count": {},
    "total_status_code_count": {
        "200": 15
    },
    "count": 0,
    "total_count": 15,
    "total_response_time": "9.947ms",
    "total_response_time_sec": 0.009947,
    "total_response_size": 2115,
    "average_response_size": 141,
    "average_response_time": "663.133µs",
    "average_response_time_sec": 0.000663133,
    "total_metrics_counts": {},
    "average_metrics_timers": {}
}
```

PR #134, opened 2019-01-10
Parallelizes all make targets

PradyumnaShome:


ASankaran:
The travis vms only have two cores so using the `-j` flag without an argument starts up too many processes. We want to limit it to 2 or 4 parallel builds at most.

ASankaran:
It seems like you cannot have two go builds which access the same files running at the same time.

See https://travis-ci.com/HackIllinois/api/builds/96943076 for the error

vscode-go had a similar issue https://github.com/Microsoft/vscode-go/issues/1868

ASankaran:
Unfortunately we'll have race conditions as long as there more that one job running, so reducing it doesn't help.

PradyumnaShome:
Fair enough. Additionally, there seemed to be a negligible effect on build time. The time spent compiling the API is dwarfed by that spent booting the VM and installing dependencies.

Issue #137, opened 2019-01-10
Allow "Mentor" and "Sponsor" roles to checkin without rsvp


ASankaran:
We don't want to have to make mentors and sponsors rsvp ahead of time. So the checkin service should let them check in without an rsvp. Modifying the `CanUserCheckin` function should be all that is required.

https://github.com/HackIllinois/api/blob/4456b5502be7e4d1a7f758d25afb101a881543ff/services/checkin/service/checkin_service.go#L75

- [x] Allow users with role "Mentor" or "Sponsor" to check in without an rsvp

PR #142, opened 2019-01-13
Allows sponsors and mentors to check-in without RSVP

PradyumnaShome:
Addresses #137.

### DIFF
--- a/services/checkin/models/user_roles.go
+++ b/services/checkin/models/user_roles.go
@@ -1,0 +1,6 @@
+package models
+
+type UserRoles struct {
+	ID    string   `bson:"id"    json:"id"`
+	Roles []string `bson:"roles" json:"roles"`
+}
--- a/services/checkin/models/user_roles.go
+++ b/services/checkin/models/user_roles.go
@@ -1,6 +1,6 @@
 package models
 
 type UserRoles struct {
-	ID    string   `bson:"id"    json:"id"`
-	Roles []string `bson:"roles" json:"roles"`
+	ID    string   `json:"id"`
+	Roles []string `json:"roles"`
 }
--- a/services/checkin/service/auth_service.go
+++ b/services/checkin/service/auth_service.go
@@ -32,7 +32,7 @@ func AddAttendeeRole(id string) error {
 */
 func GetRoles(id string) (*models.UserRoles, error) {
 	var user_roles models.UserRoles
-	status, err := apirequest.Get(config.AUTH_SERVICE+"/auth/roles/"+id+"/", user_roles)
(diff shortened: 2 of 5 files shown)


services/checkin/models/user_roles.go
```diff
@@ -0,0 +1,6 @@
+package models
+
+type UserRoles struct {
+	ID    string   `bson:"id"    json:"id"`
```
ASankaran: We shouldn't have `bson` tags anywhere in the codebase anymore. If there's somewhere they should be removed.

services/checkin/service/checkin_service.go
```diff
@@ -69,8 +70,14 @@ func UpdateUserCheckin(id string, user_checkin models.UserCheckin) error {
 	return err
 }
 
+const (
+	SponsorRole = "Sponsor"
```
ASankaran: Copy over this file https://github.com/HackIllinois/api/blob/master/services/auth/models/roles.go into the models folder for this service and use that.

services/checkin/service/auth_service.go
```diff
@@ -26,3 +26,21 @@ func AddAttendeeRole(id string) error {
 
 	return nil
 }
+
+/*
+	Gets the roles for a user given id.
+*/
+func GetRoles(id string) (*models.UserRoles, error) {
+	var user_roles models.UserRoles
+	status, err := apirequest.Get(config.AUTH_SERVICE+"/auth/roles/get/"+id+"/", user_roles)
```
ASankaran: Did you test this? I don't see any way this could work.

services/checkin/service/checkin_service.go
```diff
@@ -79,8 +86,16 @@ func CanUserCheckin(id string, user_has_override bool) (bool, error) {
 		return false, err
 	}
 
+	user_roles, err := GetRoles(id)
+
+	if err != nil {
+		panic(err)
```
ASankaran: Don't panic in service functions.

services/checkin/service/auth_service.go
```diff
@@ -26,3 +26,21 @@ func AddAttendeeRole(id string) error {
 
 	return nil
 }
+
+/*
+	Gets the roles for a user given id.
+*/
+func GetRoles(id string) (*models.UserRoles, error) {
+	var user_roles models.UserRoles
+	status, err := apirequest.Get(config.AUTH_SERVICE+"/auth/roles/"+id+"/", user_roles)
```
ASankaran: Did you test this? You should have to pass `user_roles` as a pointer. So `&user_roles`.

PradyumnaShome:
Was fixing the docs so I test with/without registration. I'll let you know once I've tested.

PR #150, opened 2019-01-17
Go formatting fixes

PradyumnaShome:
Ben and I had nothing else to do at the meeting other than plan the food distribution service, as per our meeting with @YashoSharma. We had cookies, and here's a pull request for you.

Cheers
HackIllinois Back-end Systems

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ run:
 
 .PHONY: fmt
 fmt:
-	go fmt ./...
+	gofmt -s -w .
 
 .PHONY: deploy
 deploy:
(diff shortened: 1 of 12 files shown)


ASankaran:
If we're using the `-s` flag on go fmt lets add it to the makefile and travis build.

ASankaran:
Also if we're fixing go report card errors, lets fix all three inefficient assignment warnings.

Makefile
```diff
@@ -22,7 +22,7 @@ run:
 
 .PHONY: fmt
 fmt:
-	go fmt ./...
+	gofmt -s -w .
```
ASankaran: `-w` silences the output right? It's useful for us to know which files are modified when running this.

PR #170, opened 2019-02-14
Add ability to subscribe users to topics corresponding with roles

benpankow:
Adds `POST` endpoint `/notifications/update/` that updates a user's topic subscriptions to match their roles.

### DIFF
--- a/config/dev_config.json
+++ b/config/dev_config.json
@@ -92,9 +92,9 @@
                 "Staff": "Staff",
                 "Mentor": "Mentor",
                 "Applicant": "Applicant",
-                "Attendee", "Attendee",
-                "User", "User",
-                "Sponsor", "Sponsor"
+                "Attendee": "Attendee",
+                "User": "User",
+                "Sponsor": "Sponsor"
         },
 
 
--- a/config/dev_config.json
+++ b/config/dev_config.json
@@ -88,7 +88,7 @@
 	},
 
         "GROUP_TOPIC_MAP": {
-                "Admin": "Admins"
+                "Admin": "Admin"
         },
 
 	"REGISTRATION_STAT_FIELDS": [
--- a/config/dev_config.json
+++ b/config/dev_config.json
@@ -88,8 +88,15 @@
 	},
 
         "GROUP_TOPIC_MAP": {
-                "Admin": "Admin"
+                "Admin": "Admin",
+                "Staff": "Staff",
+                "Mentor": "Mentor",
+                "Applicant": "Applicant",
+                "Attendee", "Attendee",
+                "User", "User",
+                "Sponsor", "Sponsor"
         },
+
 
 	"REGISTRATION_STAT_FIELDS": [
 		"major",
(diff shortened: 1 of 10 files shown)


config/dev_config.json
```diff
@@ -87,6 +87,17 @@
 		"event": "http://localhost:8010/event/internal/stats/"
 	},
 
+        "GROUP_TOPIC_MAP": {
+                "Admin": "Admin",
+                "Staff": "Staff",
+                "Mentor": "Mentor",
+                "Applicant": "Applicant",
+                "Attendee", "Attendee",
```
ASankaran: Invalid json.

config/dev_config.json
```diff
@@ -87,6 +87,17 @@
...
+                "User", "User",
+                "Sponsor", "Sponsor"
```
ASankaran: Invalid json.

config/production_config.json
```diff
@@ -62,6 +62,17 @@
 		"event": "http://event.api.:8010/event/internal/stats/"
 	},
 
+        "GROUP_TOPIC_MAP": {
+                "Admin": "Admin",
+                "Staff": "Staff",
+                "Mentor": "Mentor",
+                "Applicant": "Applicant",
+                "Attendee", "Attendee",
```
ASankaran: Invalid json.

gateway/services/notifications.go
```diff
@@ -75,6 +75,12 @@ var NotificationsRoutes = arbor.RouteCollection{
 		"/notifications/{id}/info/",
 		alice.New(middleware.IdentificationMiddleware, middleware.AuthMiddleware([]models.Role{models.AdminRole})).ThenFunc(GetTopicInfo).ServeHTTP,
 	},
+	arbor.Route{
```
ASankaran: This route will never be matched since `/notifications/{id}/` will match it first. So this need to be moved above that.

services/notifications/service/auth_service.go
```diff
@@ -0,0 +1,27 @@
+package service
+
+import (
+	"errors"
+	"github.com/HackIllinois/api/common/apirequest"
+	"github.com/HackIllinois/api/services/checkin/config"
+	"github.com/HackIllinois/api/services/checkin/models"
```
ASankaran: Don't have cross service imports. The notifications config should be used instead of the checkin config and the required models should be copied into the notification service's model folder.

services/notifications/service/notifications_service.go
```diff
@@ -219,6 +219,50 @@ func GetNotificationsForTopic(topic_name string) (*models.NotificationList, erro
 	return &notifications_list, nil
 }
 
+/*
+   Subscribes a user to topics corresponding to their roles, and unsubscribes a user from all other topics
+*/
+func UpdateUserSubscriptions(user_id string) (*models.TopicList, error) {
```
ASankaran: Can we have a basic test for this?

PR #191, opened 2019-02-25
Fix API docs URL in README

cktang88:


PradyumnaShome:
Thank you for the contribution @cktang88!

Issue #213, opened 2019-04-13
Remove the panic logic used to return non-200 responses


ASankaran:
Using panics is a really simple way for us to immediately abort processing a request and return an error message, but it is very non-go like in terms of panic usage. Additionally since panics unwind the stack the request never goes through our health middleware and we can't track the request.

I'd like to see blocks of code such as:
```
if err != nil {
	panic(errors.DatabaseError(err.Error(), "Could not fetch user info by ID."))
}
```
turned into something like:
```
if err != nil {
	errors.WriteError(w, errors.DatabaseError(err.Error(), "Could not fetch user info by ID."))
	return
}
```

- [ ] Remove panic logic for returning non-200 responses

PR #219, opened 2019-06-13
Remove panic logic from services

benpankow:
Addresses #213. We now directly handle errors instead of panicking and relying on the errors middleware to defer/recover. This lets us maintain the flow of requests through the API and ensures that middleware can correctly handle the request on the way out.

- [x] TODO: Remove [errors middleware](https://github.com/HackIllinois/api/blob/master/common/middleware/error.go), which should no longer be needed.


ASankaran:
You might have a plan for this already, but the logging logic from the errors middleware should be moved into the WriteError function somehow. Otherwise debugging issues in prod will be near impossible.

common/errors/errors.go
```diff
@@ -14,3 +21,38 @@ type ApiError struct {
...
+
+	if err != nil {
+		fmt.Printf("Failed to marshal error for id: %v\n", id)
+		return
+	}
+
+	fmt.Printf("ERROR: %v\n", string(error_log_message))
+}
+
+// Writes the given error to the passed HTTP response
```
bstrauch24: Code is descriptive enough; don't need this comment

common/errors/errors.go
```diff
@@ -14,3 +21,38 @@ type ApiError struct {
...
+	Error interface{}
+}
+
+func LogError(id string, error_message interface{}) {
+	log_entry := ErrorLogEntry{
+		ID:    id,
+		Error: error_message,
+	}
+
+	error_log_message, err := json.MarshalIndent(log_entry, "", "    ")
```
bstrauch24: Hard to tell what this code is doing without looking up what `MarshalIndent` means; maybe put the "tab" string in a constant called `INDENT`?

services/auth/controller/controller.go
```diff
@@ -41,13 +41,15 @@ func Authorize(w http.ResponseWriter, r *http.Request) {
 	oauth_provider, err := service.GetOAuthProvider(provider)
 
 	if err != nil {
-		panic(errors.MalformedRequestError(err.Error(), "Invalid OAuth provider."))
+		errors.WriteError(w, r, errors.MalformedRequestError(err.Error(), "Invalid OAuth provider."))
```
bstrauch24: Hopefully you used a regex for all of these!

common/apiserver/apiserver.go
```diff
@@ -19,7 +19,6 @@ func StartServer(address string, router *mux.Router, name string, initialize fun
 		return err
 	}
 
-	router.Use(middleware.ErrorMiddleware)
```
bstrauch24: Do we still need the code in `api/common/middleware/error.go`?

bstrauch24: Never mind, I think you deleted it. GitHub doesn't show that very well

PR #227, opened 2019-08-20
Merge staging into master

bstrauch24:
Pushing this through right now so resumes are uploaded to a new S3 bucket


PradyumnaShome:
If you don't want everything to go over, you could just cherry-pick the commits related to resumes.

bstrauch24:
There's no reason not to merge everything :man_shrugging:

Issue #229, opened 2019-09-05
Keep track of event check-in times


benpankow:
Right now we simply track whether or not a user has checked in to an event - we don't store any information on when this happened. It might be useful to have this information on file, especially for food check-ins.

PR #241, opened 2019-09-19
Keep track of event checkin times (#229)

michael-kwan:
Referencing #229 :

For each event and user, a list of users checked in and events checked into are kept track of in the form of an array of ids. In order to keep track of checkin times, the unix timestamp at checkin is placed into a map with each event/user id.


bstrauch24:
Looks good! Make sure you merge into the staging branch instead of the master branch :smile:

services/event/tests/event_test.go
```diff
@@ -430,7 +430,7 @@ func TestMarkUserAsAttendingEventService(t *testing.T) {
 
 	expected_event_tracker := models.EventTracker{
 		EventID: "testid",
-		Users:   []string{"testuser"},
+		Users:   map[string]int{"testuser": int(time.Now().Unix())},
```
benpankow: It looks like these test may be flaky - if they run on the boundary of a particular second, the time in the expected event may not be identical to the time in the event you create.

I would recommend allowing some tolerance in the time associated with a user's check-in. You may need to use something other than `DeepEqual` in this case.

Issue #236, opened 2019-09-14
Allow notification platforms/ARNs to be specified in config file


benpankow:
At the moment, we hardcode the two notifications platforms we support (iOS and Android) in the [service code](https://github.com/HackIllinois/api/blob/master/services/notifications/service/notifications_service.go#L315). Since we load the ARN for each service from the config anyway, it would make sense to have the config contain a mapping of all valid platform names to their ARNs:

```
"ANDROID_PLATFORM_ARN": "abc",
"IOS_PLATFORM_ARN": "def",
```

to 

```
"PLATFORM_ARNS": {
    "android": "abc",
    "ios": "def"
}
```

PR #245, opened 2019-09-22
Allow notifications/arns to be specified in config file (Issue #236)

nitishp1812:
Edited the config json file to contain the platform arns for all platforms. Parsed the platform arns into a map which was then used in the notifications_service.go file to set the platform_arn variable


gateway/services/services.go
```diff
@@ -38,7 +39,7 @@ var Routes = arbor.RouteCollection{
 }
 
 func Gateway(w http.ResponseWriter, r *http.Request) {
-	fmt.Fprintf(w, "The API Gateway Lives")
+	fmt.Fprintf(w, "The API Gateway Lives (found by the king)")
```
benpankow: Make sure to remove this change from your PR :laughing:

nitishp1812:
Restored original gateway message

Issue #235, opened 2019-09-14
Add endpoint to unregister a given device for a user


benpankow:
Right now, we can only remove a user's devices both from our databases and from Amazon SNS manually. It would be good to be able to unregister a device given its token and platform, or its ARN.

It looks like [CreatePlatformEndpoint](https://docs.aws.amazon.com/sns/latest/api/API_CreatePlatformEndpoint.html), which we use to register a device, can also be used to obtain the ARN of an already registered device, given its token and platform.

thomaslanghorst:
If noone is working on this issue I would like to contribute.

PR #248, opened 2019-09-23
Add endpoint to unregister a given device for a user (#235)

akulio4668:
Allows us to unregister a user using their token and platform, or their ARN, instead of having to do it manually

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,3 +1,0 @@
-
-# Default ignored files
-/workspace.xml
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,3 @@
+
+# Default ignored files
+/workspace.xml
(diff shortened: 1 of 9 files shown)


.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+
```
benpankow: Can you remove these files from the commit?

services/notifications/service/notifications_service.go
```diff
@@ -355,6 +355,60 @@ func RegisterDeviceToUser(token string, platform string, id string) error {
...
+
+	devices, err := GetUserDevices(id)
+
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < len(devices); i++ {
+		if devices[i] == device_arn {
+			devices = append(devices[:i], devices[i+1:]...)
```
benpankow: It would be good if we could write a test for this functionality. Since tests have `IS_PRODUCTION` false, we should be able to test the db functionality fine.

akulio4668: Definitely. Do we just write any tests for our program in the /gateway/tests/gateway_test.go or is there another place where we write them?

benpankow: Tests go in the specific `/services/notifications/service/tests/` folder

Issue #231, opened 2019-09-12
Add event filter endpoint


benpankow:
Currently, we have filtering endpoints for a few different services (`GET /user/filter`, for example). It would be nice to do something similar for events.

`GET /event/filter/?sponsor=Fulcrum`

In addition, searching within a particular time window would be helpful.

Something like:
`GET /event/filter/?minStartTime=1532202702&maxStartTime=1532212702`
 or
`GET /event/filter/?startTimeGt=1532202702`

- [x] Basic functionality
- [ ] Time range filtering

benpankow:
@aryaman4

aryaman4:
@benpankow I'm done with the code but testing on local is giving me a 403 forbidden. I'm guessing I'm supposed to add the admin token to the request but I'm not sure about the auth field which I'm supposed to use for postman.

aryaman4:
I'm also not sure about how the mongo is set up structurally, so if there's any docs or if you could briefly explain, I'll set up a dummy on local for better testing.

PR #249, opened 2019-09-23
Event filter (Issue #231)

aryaman4:
Created endpoint for event filtering. Issue #231


aryaman4:
@benpankow I'm not sure what the problem is since it builds locally

bstrauch24:
Make sure you run "go fmt" before committing!

Issue #242, opened 2019-09-20
Add blobstore delete endpoint


benpankow:
We currently have PUT, GET, and POST endpoints for the blobstore (https://docs.hackillinois.org/reference/services/Upload/). While our current use cases only include certain time constants, if we decide to use the blobstore for some sort of CRUD functionality, it would be good to be able to call `DELETE /upload/blobstore/exampleblob/` to remove the entry w/ id `exampleblob`.

Look to the event `DELETE` endpoint as reference.

thomaslanghorst:
Is this issue still open for contribution? If so, I would like to help out.

benpankow:
There's already a PR made for this issue (#254), but if you're interested in any of the other open issues, let me know!

thomaslanghorst:
Oh, ok. I saw (#250) as closed but not merged. But I did not check the open PRs. I am new to open source. Thanks for helping me out. I will look for another open issue to contribute to.

PR #254, opened 2019-10-03
Issue #242: Add blobstore delete endpoint

hsiehby:
I added the endpoint deleteBlob for deleting blobstores with the given id. Also, the documentation has also been modified to include this endpoint.

### DIFF
--- a/gateway/services/upload.go
+++ b/gateway/services/upload.go
@@ -1,12 +1,13 @@
 package services
 
 import (
+	"net/http"
+
 	"github.com/HackIllinois/api/gateway/config"
 	"github.com/HackIllinois/api/gateway/middleware"
 	"github.com/HackIllinois/api/gateway/models"
 	"github.com/arbor-dev/arbor"
 	"github.com/justinas/alice"
-	"net/http"
 )
 
 const UploadFormat string = "RAW"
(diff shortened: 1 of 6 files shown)


gateway/services/upload.go
```diff
@@ -49,6 +50,12 @@ var UploadRoutes = arbor.RouteCollection{
 		"/upload/blobstore/{id}/",
 		alice.New(middleware.IdentificationMiddleware).ThenFunc(GetBlob).ServeHTTP,
 	},
+	arbor.Route{
+		"DeleteBlob",
+		"DELETE",
+		"/upload/blobstore/{name}/", // ?? Is it correct to use {name}?
```
benpankow: `name` should work fine, but I would go with `id` since that's what we use elsewhere in this endpoint and what you use in the controller code

go.mod
```diff
@@ -15,3 +15,5 @@ require (
 	gopkg.in/go-playground/validator.v9 v9.25.0
 	gopkg.in/mgo.v2 v2.0.0-20180705113604-9856a29383ce
 )
+
+go 1.13
```
benpankow: was this change intentional/needed?

hsiehby: I found that this line is automatically added whenever I run the test.
I have removed it.

gateway/services/upload.go
hsiehby: Okay, I have changed it to id.

hsiehby:
I have removed unnecessary lines and added a test case. Please review my newer commit, thanks!

PR #257, opened 2019-10-10
Add basic endpoint to filter events

benpankow:
Addresses #231. Adds the `/event/filter/` endpoint, which lets us filter events by particular criteria. Not restricted to any particular user group, since we'll probably use this for the dashboard/logged-out app view.

Usage is of the form `/event/filter/?sponsor=Example%20sponsor&eventType=WORKSHOP`.


benpankow:
It would also be good to be able to query less than/greater than particular timestamps, which I'll address in a follow-up PR.

Issue #240, opened 2019-09-19
Improve type assertion error messages


benpankow:
When we try to convert certain input data types into types matching a datastore schema, we throw a generic "Type mismatch in data and definition" error. 

This makes it difficult to work on the Registration endpoint - since the Registration model is loaded from our config file, we can't look to a model file as a source of truth. It would be good to include the expected type in the error we throw in https://github.com/HackIllinois/api/blob/master/common/datastore/conversions.go.

PR #260, opened 2019-10-30
Improve type assertion errors for datastore

benpankow:
Addresses #240. If decomposing JSON/BSON into a datastore fails because the input data doesn't match the types specified by the definition, the returned error messages will specify which field failed, which type was expected, and the type provided.

Adds a small test.

Example output:
```
{
    "status": 500,
    "type": "INTERNAL_ERROR",
    "message": "Could not decode user registration information. Possible failure in JSON validation, or invalid registration format.",
    "raw_error": "Error in field firstName: Type mismatch in data and definition. Expected string, got float64"
}
```

### DIFF
--- a/common/datastore/conversions.go
+++ b/common/datastore/conversions.go
@@ -4,7 +4,7 @@ func toInt(raw_data interface{}, definition DataStoreDefinition) (interface{}, e
 	data, ok := raw_data.(float64)
 
 	if !ok {
-		return nil, ErrTypeMismatch
+		return nil, NewErrTypeMismatch(raw_data, "float64")
 	}
 
 	return int64(data), nil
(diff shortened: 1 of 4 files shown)


common/datastore/conversions.go
```diff
@@ -44,7 +44,7 @@ func toObject(raw_data interface{}, definition DataStoreDefinition) (interface{}
 	unfiltered_data, ok := raw_data.(map[string]interface{})
 
 	if !ok {
-		return nil, ErrTypeMismatch
+		return nil, NewErrTypeMismatch(raw_data, "[]interface{}")
```
bstrauch24: Should this be "map[string]interface{}" instead of "[]interface{}"?

benpankow: Good catch, thanks

Issue #224, opened 2019-08-08
Migrate LinkedIn to API v2


bstrauch24:
LinkedIn no longer supports v1 of their API

benpankow:
This could be a good intro task. See https://developer.linkedin.com/docs/v2/oauth2-client-credentials-flow#hero-par_productbanner for up-to-date LinkedIn OAuth docs.

PR #264, opened 2019-11-07
Migrate to LinkedIn API v2 #224

hsiehby:
I mainly modified the `getUserInfo` function, which is responsible for retrieving user's name, id, and email address. Now a separate request is required to get the email address. Some models are added according to the new API response format. In `GetAuthorizationRedirect` function, I changed the scope from `r_basicprofile` to `r_liteprofile` since basic profile access is now restricted to their enterprise plan users.


benpankow:
Looks good - build looks like it's failing because of formatting reasons. Probably just need to run `make fmt`.

Issue #263, opened 2019-10-31
Delete Production Config


brianstrauch:

bstrauch24:
The production config now lives in a separate private repo

brianstrauch:
@heesooy JSYK looks the prod config is still in this repo and you should probably delete it: https://github.com/HackIllinois/api/blob/staging/config/production_config.json

heesooy:
😳 thanks for double checking!

PR #266, opened 2019-11-13
Refactor production config

heesoo4199:
Addresses #263 

1. Removed production_config.json from this repo
2. Updated registration object according to new specification in Google docs
3. Updated registration documentation to reflect new registration object

*The registration object will definitely need revision on the validation rules.

Based on manual testing with postman, the registration service seemed to work fine. Let me know if you know/see any problems with the changes.


heesoo4199:
The checks are not passing because `make test` is failing. One of the issues is that the tests use a registration field that was removed.

I fixed this, but now I'm running into another problem: even though the registration config does not require the "linkedin" field, if `user_registration_data` does not have a value for "linkedin", then `CreateMentorRegistration` fails, saying

> Key 'linkedin' with value '' failed validation 'required'

I made no changes to the mentor registration object or any tests related to them.

@benpankow do you know what could be causing this?

ASankaran:
I think this was my fault. I must have introduced this issue when first setting up the tests, but the mentor registration was a subset of the user registration so it worked.

https://github.com/HackIllinois/api/blob/2533c60c3cd0b0fe8653561c3da3cb7cad2c7175/services/registration/tests/registration_test.go#L359

That line should be `mentor_registration_data` instead of `user_registration_data`. The error should go away once that's changed.

heesoo4199:
Need to update stats fields as registration fields changed #276 

Also need to update RSVP docs

heesoo4199:
Made modifications you specified in api-config

PR #269, opened 2019-11-20
Track number of intended recipients for each notification order

### LINKED ISSUE #268: Track number of intended recipients for each notification order
@bstrauch24 and I ran into issues with the API attempting to send notifications and the success, failure counts on a particular order not updating. A change to include the number of intended recipients in an order would be helpful in diagnosing these issues.

benpankow:
Resolves #268. Adds a new field, `numRecipients`, to each notification order, which tracks the number of devices the notification is intended to be sent to.

### DIFF
--- a/services/notifications/models/notification_order.go
+++ b/services/notifications/models/notification_order.go
@@ -1,8 +1,9 @@
 package models
 
 type NotificationOrder struct {
-	ID      string `json:"id"`
-	Success int    `json:"success"`
-	Failure int    `json:"failure"`
-	Time    int64  `json:"time"`
+	ID            string `json:"id"`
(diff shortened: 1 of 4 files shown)


services/notifications/models/notification_order.go
```diff
@@ -1,8 +1,9 @@
 package models
 
 type NotificationOrder struct {
-	ID      string `json:"id"`
-	Success int    `json:"success"`
-	Failure int    `json:"failure"`
-	Time    int64  `json:"time"`
+	ID            string `json:"id"`
+	NumRecipients int    `json:"numRecipients"`
```
bstrauch24: I'd just call this recipients. Or rename the other fields to numSuccess and numFailure

PR #270, opened 2019-11-20
Added Patch method for the database.

hsiehby:
I added `Patch` method in `mongo_database.go` to enable updates of only the selected fields in an entry.
This method can be used by the patch methods in endpoints such as registration, blobstore, etc.
I also modified the `Database interface` in `database.go` to reflect the change.

### DIFF
--- a/common/database/mongo_database.go
+++ b/common/database/mongo_database.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/HackIllinois/api/common/config"
 	"gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
 )
 
 /*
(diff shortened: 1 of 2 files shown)


common/database/mongo_database.go
```diff
@@ -205,6 +206,20 @@ func (db *MongoDatabase) UpdateAll(collection_name string, selector interface{},
 	return &change_results, convertMgoError(err)
 }
 
+/*
+	Finds an item based on the given selector and patches it with the data in update
+*/
+func (db *MongoDatabase) Patch(collection_name string, selector interface{}, update *map[string]interface{}) error {
```
benpankow: If it's not a lot of extra effort, it would be nice to have `update` be an `interface{}` to match the signature of the other database functions. Not a big deal if not.

common/database/mongo_database.go
```diff
@@ -205,6 +206,20 @@ func (db *MongoDatabase) UpdateAll(collection_name string, selector interface{},
...
+	current_session := db.GetSession()
+	defer current_session.Close()
+
+	collection := current_session.DB(db.name).C(collection_name)
+
+	// Use mongodb set operator to update only the provided fields.
+	err := collection.Update(selector, bson.M{"$set": update})
```
benpankow: It's my understanding that this will only merge top-level changes, not nested fields. If so, it might be worth noting this in the comment above.

ie

```
{
  "foo": {
    "bar1": 1,
  }
}
```

merged with

```
{
  "foo": {
    "bar2": 2,
  }
}
```

results in

```
{
  "foo": {
    "bar2": 2,
  }
}
```

PR #273, opened 2019-11-21
Add helper code to create db queries from URL parameters

benpankow:
Our filtering endpoints (`/event/filter` and `/user/filter`) interpret all URL parameters as strings, and as such filtering on fields which hold integers (such as start times) doesn't work as expected.

This PR introduces a helper function to the `database` package which correctly casts parameters to the expected type. It does this by taking in a copy of the struct which models the table we're querying, and uses go's `reflection` package to figure out what each field's type is and what its name is in the database.

A later PR will actually switch over existing filtering endpoints to use this function, and can introduce additional functionality (pagination, range queries, sorting).

I was a bit concerned about `reflection`'s impact on performance - there might be a better way to do this, let me know if so. Below are some benchmarks which compare how `/event/filter` compares using old filtering and the new method. There's only a ~3% performance hit. 

image

Currently, this code supports `string` (unchanged), `int64`, and `bool`.

### DIFF
--- a/common/database/filter.go
+++ b/common/database/filter.go
@@ -11,7 +11,7 @@ import (
 	Returns a map for the given model, where each key is the JSON field name and
 	each value is a string representation of that field's type.
 
-	For example, a struct which holds a string Name value has types["nane"] = "string"
+	For example, a struct which holds a string Name value has types["name"] = "string"
 */
 func GetFieldTypes(model interface{}) map[string]string {
 	expected_types := make(map[string]string)
--- a/common/database/filter.go
+++ b/common/database/filter.go
@@ -1,0 +1,82 @@
+package database
+
+import (
+	"errors"
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+/*
+	Returns a map for the given model, where each key is the JSON field name and
+	each value is a string representation of that field's type.
+
+	For example, a struct which holds a string Name value has types["nane"] = "string"
+*/
+func GetFieldTypes(model interface{}) map[string]string {
+	expected_types := make(map[string]string)
+
+	v := reflect.ValueOf(model).Type()
+	for i := 0; i < v.NumField(); i++ {
+		field := v.Field(i)
+
+		json_name := strings.ToLower(field.Tag.Get("json"))
+		expected_types[json_name] = field.Type.Name()
+	}
+
+	return expected_types
+}
+
+/*
+	Turns a series of string URL parameters into a query for a particular data type
+	Returns a map of generated QuerySelectors
+*/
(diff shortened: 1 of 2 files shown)


common/database/filter.go
```diff
@@ -0,0 +1,82 @@
...
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+/*
+	Returns a map for the given model, where each key is the JSON field name and
+	each value is a string representation of that field's type.
+
+	For example, a struct which holds a string Name value has types["nane"] = "string"
```
PradyumnaShome:
```suggestion
	For example, a struct which holds a string Name value has types["name"] = "string"
```

common/tests/filter_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	query, err := database.CreateFilterQuery(params, TestStruct2{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !reflect.DeepEqual(query, expected_query) {
+		t.Errorf("Wrong order.\nExpected %v\ngot %v\n", expected_query, query)
+	}
+}
```
PradyumnaShome: How about a test for non-existent fields, and duplicate fields?

common/database/filter.go
benpankow: Good catch

common/tests/filter_test.go
benpankow: I added a non-existent field test, but maps don't support duplicate keys so I couldn't add a duplicate entry

common/tests/filter_test.go
```diff
@@ -0,0 +1,86 @@
...
+	params := map[string][]string{
+		"hackValue":     {"foo,bar"},
+		"illinoisValue": {"55,63"},
+		"testValue":     {"true,false,true"},
+	}
+
+	expected_query := map[string]interface{}{
+		"hackvalue":     database.QuerySelector{"$in": []string{"foo", "bar"}},
+		"illinoisvalue": database.QuerySelector{"$in": []int64{55, 63}},
+		"testvalue":     database.QuerySelector{"$in": []bool{true, false, true}},
```
PradyumnaShome: Just saw this - what does this logically mean? That it can be true or false?

common/tests/filter_test.go
PradyumnaShome: Yup this looks good.

common/tests/filter_test.go
benpankow: Yeah, that query is a bit nonsensical. I'll improve it in a follow-up PR when additional functionality is added.

PR #293, opened 2020-01-17
Support ctrl-c on Mac

### LINKED ISSUE #292: Ctrl-C after make run doesn't stop services on Mac

bstrauch24:
Closes #292

### DIFF
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -3,6 +3,7 @@
 trap cleanup INT
 
 function cleanup {
+  echo "Stopping services"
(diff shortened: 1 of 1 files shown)


scripts/run.sh
```diff
@@ -3,7 +3,6 @@
 trap cleanup INT
 
 function cleanup {
-	echo "Cleaning up services"
```
ASankaran: I'd leave this in just so that users know that their ctrl-c has been received and that if the process hasn't exited it is because kill is taking a bit longer an expected.

PR #298, opened 2020-01-22
Add less than, greater than, not in filter operators

benpankow:
Adds some additional filter operators (less than, greater than, not in) which can be used for range queries.

These are used by attaching suffixes `Lt`, `Gt`, and `Not` to field names, such as:
`/user/filter/?usernameGt=abcde`

### DIFF
--- a/common/database/filter.go
+++ b/common/database/filter.go
@@ -7,6 +7,15 @@ import (
 	"strings"
 )
 
+type QueryType string
+
+const (
+	LessThan    QueryType = "LessThan"
+	In          QueryType = "In"
+	GreaterThan QueryType = "GreaterThan"
+	NotIn       QueryType = "NotIn"
+)
+
 /*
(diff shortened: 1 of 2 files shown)


common/database/filter.go
```diff
@@ -27,6 +36,66 @@ func GetFieldTypes(model interface{}) map[string]string {
...
+		return nil, errors.New("Invalid operation on booleans")
+	}
+	return qs, nil
+}
+
+func ParseQueryType(key string) (QueryType, string) {
+	query_type := In
+
+	if strings.HasSuffix(key, "Lt") {
+		key = key[0 : len(key)-2]
```
ASankaran: Is it possible for `len(key)` to be less than 2? (Or 3 below?)

benpankow: If the string has the suffix `Lt` it should have length >= 2. I think this should be okay since we only slice the end off the keys if they have the corresponding suffixes.

PR #305, opened 2020-02-03
Add tags field to event locations

benpankow:
Adds a `tags` field to event locations, which contains information about building/floor which can be used to display the proper map for each event.

Tags can be any of the following: `SIEBEL0`, `SIEBEL1`, `SIEBEL2`, `SIEBEL3`, `SIEBEL4`, `ECEB1`, `ECEB2`, `ECEB3`, `ECEB4`, `ECEB5`, `KENNEY`, `DCL`

Some of these might be unnecessary, unsure how much of each building we host events in

### DIFF
--- a/services/event/models/event.go
+++ b/services/event/models/event.go
@@ -13,7 +13,7 @@ type Event struct {
 
 type EventLocation struct {
 	Description string   `json:"description" validate:"required"`
-	Tags        []string `json:"tags"        validate:"required,dive,oneof=SIEBEL0 SIEBEL1 SIEBEL2 SIEBEL3 SIEBEL4 ECEB1 ECEB2 ECEB3 ECEB4 ECEB5 KENNEY DCL"`
+	Tags        []string `json:"tags"        validate:"required"`
 	Latitude    float64  `json:"latitude"    validate:"required"`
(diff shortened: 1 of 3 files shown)


services/event/models/event.go
```diff
@@ -12,7 +12,8 @@ type Event struct {
 }
 
 type EventLocation struct {
-	Description string  `json:"description" validate:"required"`
-	Latitude    float64 `json:"latitude"    validate:"required"`
-	Longitude   float64 `json:"longitude"   validate:"required"`
+	Description string   `json:"description" validate:"required"`
+	Tags        []string `json:"tags"        validate:"required,dive,oneof=SIEBEL0 SIEBEL1 SIEBEL2 SIEBEL3 SIEBEL4 ECEB1 ECEB2 ECEB3 ECEB4 ECEB5 KENNEY DCL"`
```
ASankaran: As an FYI. This starts to break down the API as a generic solution for events. RP, for example, may need a different set of tags. I'd recommend reading this in from a config file.

bstrauch24: Alternatively, we could just keep it simple and remove the `oneof=...` from the validation entirely

Issue #261, opened 2019-10-31
Add pagination and sorting parameters to users filter endpoint


heesoo4199:
We want to paginate the response for the GET users endpoint, as well as sort the result by some key. We should modify the /user/filter/ endpoint to support these features. For sorting, we should support sorting by multiple keys for tiebreaking.

The request should be in some form like `/user/filter/?key=value&sortby=firstName,lastName&limit=5&p=2` to sort firstName then lastName, limit to 5 users per page, and get the 2nd page.

GabrielRato:
Hello @heesoo4199 , do you still need help on this? I could take a look 😄

PR #353, opened 2020-11-05
Feat/user filtering sortby

mataslauzadis:
#261 

Adds sorting functionality to the /user/filter/ service.

To use it, provide a comma-separated "sortby" parameter.

`/user/filter/?key=value&sortby=FirstName,LastName`

To reverse the sort, add a minus (-) to the desired field. An example is provided in the service documentation.

`/user/filter/?key=value&sortby=-FirstName,LastName`

### DIFF
--- a/services/user/service/user_service.go
+++ b/services/user/service/user_service.go
@@ -3,6 +3,7 @@ package service
 import (
 	"errors"
 	"net/url"
+	"strings"
 
 	"github.com/HackIllinois/api/common/database"
 	"github.com/HackIllinois/api/services/user/config"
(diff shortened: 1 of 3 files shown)


services/user/service/user_service.go
```diff
@@ -67,14 +68,45 @@ func SetUserInfo(id string, user_info models.UserInfo) error {
...
+
+		// Create and fill the sort fields
+		var sort_fields []database.SortField
+
+		for _, field := range sort_parameters {
+			// Push to lowercase because MongoDB columns are all lowercase
+			field = strings.ToLower(field)
+			field = strings.TrimSpace(field)
+
+			// TODO Add a check here to make sure the requested sort field is actually a field in the model.
```
heesoo4199, line 96: Good idea! We should probably do this for all filter & sort params for other services too.

services/user/tests/user_test.go
```diff
@@ -43,7 +44,7 @@ func TestMain(m *testing.M) {
 }
 
 /*
-	Initialize databse with test user info
+	Initialize database with test user info
```
heesoo4199, line 47: :thumbsup:

heesoo4199 (approved):
Looks good! Nice testing. Good point about checking filter & sort params to see if they are valid keys in the db. I wonder if mongo already ignores nonexistent keys, but it would be nice to return an error for QoL.

Issue #354, opened 2020-11-05
Add pagination to users filter endpoint


mataslauzadis:
We want to paginate the response for the GET users endpoint. We should modify the /user/filter/ endpoint to support this feature.

The request should be in some form like /user/filter/?key=value&sortby=firstName,lastName&limit=5&p=2 to sort firstName then lastName, limit to 5 users per page, and get the 2nd page.

PR #357, opened 2020-11-19
Feat/user filtering pagination

mataslauzadis:
#354 

Adds pagination to the /user/filter endpoint.

Parameter "p": defines the page number you are requesting
Parameter "limit": defines how many elements to display per page

For example, the following request: `/user/filter/?key=value&p=1&limit=5` will return the first 5 Users (index 0 through 4).

If the requested page larger than the list of Users, it will be truncated.


heesoo4199 (approved):
looks good

heesoo4199 (approved):

PR #360, opened 2021-02-17
Remove travis and test github actions

heesoo4199:
1. Removed references to Travis in unit tests and readmes
2. Added simple bug that should be caught by the build, to be corrected.


PatrickKan (approved):
free github actions :)

PR #361, opened 2021-02-17
Add additional Profile endpoints

mataslauzadis:
- GetProfileById (uses GetProfile)
- GetFilteredProfiles
- GetProfileLeaderboard


services/profile/service/profile_service.go
```diff
@@ -142,3 +143,85 @@ func GetAllProfiles() (*models.ProfileList, error) {
...
+		Profiles: profiles,
+	}
+
+	return &profile_list, nil
+}
+
+/*
+	Returns a list of profiles filtered upon teamStatus and interests. Will be limited to only include the first "limit" results.
+*/
+func GetFilteredProfiles(teamStatus string, interests_string string, limit int) (*models.ProfileList, error) {
```
heesoo4199, line 179: Could we move the filtering logic into mongodb as much as possible?

Our existing database interface (/common/database) should be able to handle the `teamStatus` and `interests` queries. Sorting by matching interests will be trickier and isn't supported by our database interface AFAIK.

services/profile/controller/controller.go
```diff
@@ -37,6 +42,24 @@ func GetProfile(w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(user_profile)
 }
 
+/*
+	GetProfileById is used to get a profile for a provided id.
+*/
+func GetProfileById(w http.ResponseWriter, r *http.Request) {
+	parameters := r.URL.Query()
+
+	id := parameters.Get("id")
```
heesoo4199, line 51: This should work, but for consistency with the other services, let's change this to

```id := mux.Vars(r)["id"]```

**Review by heesoo4199 — CHANGES_REQUESTED** [review]
Hey Matas, other than a few minor things, everything looks good! If you could make those changes, it should be good to merge.

**Comment by heesooy** [comment]
I decided to pick up where Matas left off, since we need these endpoints out soon.

Main changes since then are:

1. Refactor /profile/controller/controller.go to move logic/error handling into the profile_service.go file
2. Move filtering for `teamStatus` and `interests` for the `/profile/search/` endpoint into Mongo rather than doing it manually in go
3. Remove sorting the results for `/profile/search/` by comparing the number of common interests -- for now, for the sake of simplicity, let's just return the profiles in the order Mongo returns them. We can figure out some kind of metric for ordering later.
4. Fix a naming issue for `/common/utils/math.go:Max()` since it was actually doing a min operation

@PatrickKan (please lmk if there's anything obvious I've missed!)

File: documentation/docs/reference/services/Profile.md [hunk 3]
```diff
@@ -147,4 +167,87 @@ Response format:
...
+            "email": "john@gmail.com",
+            "github": "JSmith",
+            "linkedin": "john-smith",
+            "interests": [
+                "deep learning",
+                "ice skating",
+                "python"
+            ],
+            "points": 1000,
+        },
```
PatrickKan, line 193: Might want to only return the points and name in the payload since we dont want to expose emails. We could also use discord usernames if we want to provide more anonymity since displaying names could be kinda awk too

PatrickKan (commented):
TODO in next PR:
1. Adjust the permissions for the endpoints (everyone can access leaderboard, but only staff and admin should be able to access the others)
2. Remove id from profile struct altogether to avoid potential security/privacy concerns
3. Change leaderboard endpoint to return less information (just the name and points)

documentation/docs/reference/services/Profile.md
heesooy: Actually, good catch here. Looks like the docs here are using the wrong data fields. Let me correct this for this PR and work on the other changes in a new PR for granularity.

PatrickKan (approved):
Looks good, addressing the other issues mentioned in the comment in next PR

heesooy (approved):

PR #363, opened 2021-02-22
Delete production config

heesooy:
#263


PatrickKan (approved):
Could move the contents of the prod config repo into here instead of deleting if there arent any secrets

PR #371, opened 2021-03-10
Profile search privacy

mataslauzadis:
- Moved the existing Profile /search/ code to a new endpoint, /filtered/, only accessible by AdminRole and StaffRole
- Made a new endpoint for /search/ which returns the desired parameters, as well as filtering out the Profiles that have the TeamStatus "Not Looking" (this status can be changed to whatever we need it to be)


heesooy (approved):
LGTM

PR #375, opened 2021-03-17
Feat/less leaderboard fields

mataslauzadis:
- Allow anyone to access the /profile/leaderboard/ endpoint
- Limit return fields of the /profile/leaderboard/ endpoint to id, firstName, lastName, points

### DIFF
--- a/gateway/services/profile.go
+++ b/gateway/services/profile.go
@@ -47,7 +47,7 @@ var ProfileRoutes = arbor.RouteCollection{
 		"GetProfileLeaderboard",
 		"GET",
 		"/profile/leaderboard/",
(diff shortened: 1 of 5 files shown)


gateway/services/profile.go
```diff
@@ -47,7 +47,7 @@ var ProfileRoutes = arbor.RouteCollection{
 		"GetProfileLeaderboard",
 		"GET",
 		"/profile/leaderboard/",
-		alice.New(middleware.AuthMiddleware([]models.Role{models.AdminRole, models.AttendeeRole, models.StaffRole, models.MentorRole}), middleware.IdentificationMiddleware).ThenFunc(GetProfileLeaderboard).ServeHTTP,
+		http.HandlerFunc(GetProfileLeaderboard).ServeHTTP,
```
mataslauzadis, line 50: This line can also be changed to:
`
alice.New().ThenFunc(GetProfileLeaderboard).ServeHTTP,
`
but since we are not using middlewares I removed the instantiation of Alice.

heesooy (approved):

PR #378, opened 2021-03-22
Event Code handling

DevSatpathy:
(WIP) Creates an event code upon generation of an event, and adds an endpoint for the admin frontend portal to retrieve this code.


DevSatpathy:
TODO: Write some tests for the cross service logic, and perhaps change the way put works on the backend i.e. prevent users from directly putting points to themselves

gateway/services/event.go
```diff
@@ -85,6 +85,24 @@ var EventRoutes = arbor.RouteCollection{
...
+		"UpdateEventCode",
+		"PUT",
+		"/event/code/{id}/",
+		alice.New(middleware.AuthMiddleware([]models.Role{models.AdminRole}), middleware.IdentificationMiddleware).ThenFunc(PutEventCode).ServeHTTP,
+	},
+	arbor.Route{
+		"Checkin",
+		"GET",
+		"/event/checkin/{code}/",
+		alice.New(middleware.AuthMiddleware([]models.Role{models.AdminRole}), middleware.IdentificationMiddleware).ThenFunc(Checkin).ServeHTTP,
```
heesooy, line 104: This should be available to all the roles that have access to GET /profile/

heesooy: Might also want to make all these Admin endpoints available to Staff as well.

services/event/models/checkin_result.go
```diff
@@ -0,0 +1,7 @@
+package models
+
+type CheckinResult struct {
+	NewPoints   int    `json:"newPoints"`
```
heesooy, line 4: See my comment about lowercase camelcase above.

services/event/service/event_service.go
```diff
@@ -176,6 +176,18 @@ func CreateEvent(id string, event models.Event) error {
...
 	err = db.Insert("eventtrackers", &event_tracker)
 
+	if err != nil {
+		return err
+	}
+
+	event_code := models.EventCode{
+		ID:         id,
+		Code:       code,
+		Expiration: 1521388800,
```
heesooy, line 186: What's this hardcoded value?

services/event/service/event_service.go
```diff
@@ -461,3 +473,59 @@ func GetStats() (map[string]interface{}, error) {
...
+	}
+
+	return &eventCode, nil
+}
+
+/*
+	Updates the event code and end time with the given id
+*/
+func UpdateEventCode(id string, eventCode models.EventCode) error {
+
```
heesooy, line 523: Nitpick: no newline here

services/profile/controller/controller.go
```diff
@@ -196,3 +199,57 @@ func GetValidFilteredProfiles(w http.ResponseWriter, r *http.Request) {
...
+	event_id := mux.Vars(r)["id"]
+
+	redemption_status, err := service.EventRedemptionStatus(id, event_id)
+
+	if err != nil {
+		errors.WriteError(w, r, errors.DatabaseError(err.Error(), "Could not check if event was redeemed for id "+id+"."))
+		return
+	}
+	if !redemption_status {
+		errors.WriteError(w, r, errors.DatabaseError(err.Error(), "Event with id "+id+" was already redeemed."))
```
heesooy, line 217: Could you write a t3st or manually verify that this indeed returns an http error code? Just want to double check that we are not returning a 200 with an error message in the body, which I think we might be doing in other places in our api.

services/profile/controller/controller.go
```diff
@@ -196,3 +199,57 @@ func GetValidFilteredProfiles(w http.ResponseWriter, r *http.Request) {
...
+
+	json.NewEncoder(w).Encode(redemption_status)
+}
+
+/*
+	AwardPoints gives the specified number of points to the current user.
+*/
+func AwardPoints(w http.ResponseWriter, r *http.Request) {
+	id := r.Header.Get("HackIllinois-Identity")
+	points, err := strconv.Atoi(mux.Vars(r)["points"])
```
heesooy, line 229: This err is unused

heesooy: Again, reminder rhat this adds points to the wrong profile since only JWTs with `AdminRole` can hit this endpoint. A test to check for this would be nice.

heesooy (changes requested):
Few glaring issues and some minor things. The approach should work otherwise!

We are in need of some tests. Preferably unit tests, but manual works for now.

documentation/docs/reference/services/Event.md
```diff
@@ -361,3 +361,60 @@ Response format:
...
+
+GET /event/checkin/{code}/
+----------------------------
+
+Retrieves a struct that contains information about the event checkin status, point increment value, and total point number.
+
+Response format:
+```
+{
+    "NewPoints": 10,
```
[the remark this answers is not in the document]
DevSatpathy: Correct, the docs should have this as lowercase, as that's the way it's represented in the json.

services/event/models/checkin_result.go
DevSatpathy: The variable name of the struct should be uppercase camel case, but the json should be lower camel case (this is consistent with the casing found in the other models, i.e. the event.go model).

services/event/service/event_service.go
DevSatpathy: This is an arbitrary time value prior to the event. It can be changed to be 0 honestly (I think I will do that actually), so the initial value is the start of epoch time.

gateway/services/profile.go
```diff
@@ -61,6 +61,18 @@ var ProfileRoutes = arbor.RouteCollection{
...
 	},
+	arbor.Route{
+		"GetEventRedemptionStatus",
+		"GET",
+		"/profile/checkin/{id}/",
+		alice.New(middleware.AuthMiddleware([]models.Role{models.AdminRole}), middleware.IdentificationMiddleware).ThenFunc(GetEventRedemptionStatus).ServeHTTP,
+	},
+	arbor.Route{
+		"AwardPoints",
+		"GET",
```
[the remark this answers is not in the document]
DevSatpathy: So I discussed this with someone else, and even though we're effectively sending data to the server for this processing, I was told it's not that bad to just leave it, especially since we're retrieving some new information about the status of the checkin.

services/event/models/checkin_result.go
heesooy: Gotcha, just wanted to double check

services/event/service/event_service.go
heesooy: > This is an arbitrary time value prior to the event. It can be changed to be 0 honestly (I think I will do that actually), so the initial value is the start of epoch time.

That will make all of the codes expired by default? What if we do `event.endTime + SOME_GRACE_PERIOD`

heesooy:
Could you also add the internal endpoints to the docs? Sorry -- busy work, but other services also document internal endpoints. We can add a note under each that it's for internal api use.

You can do this later since we need to role out the feature asap.

services/profile/controller/controller.go
DevSatpathy: Refactored to not need this error (Using a post instead of a get)

services/event/service/event_service.go
DevSatpathy: Expiration is now initialized to end time.

gateway/services/profile.go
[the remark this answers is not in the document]
heesooy: Changing this to a post.

heesooy (approved):
Looks good for now. We should roll this out and work on bugfixes throughout next week while mobile implements these

PR #379, opened 2021-03-24
Profile service docs format fix

heesooy:
- Add trailing slash to GET /profile/{id}/ endpoint
- Add Profile to the site's navigation bar


mataslauzadis (approved):
Looks good!

PR #380, opened 2021-03-24
Give applicant, staff, and mentor roles access to profile service

### LINKED ISSUE #377: Allow "Applicant" role to create profiles
For HackIllinois 2021, we're going to auto-admit and RSVP will be pretty much unused, so add the "Applicant" role to the POST and PUT /profile/ endpoint so that the front end does not have to submit an RSVP and request a new token with the "Attendee" role.

heesooy:
Closes #377 

- Give `Applicant` role access to profile service methods, for HackIllinois 2021
- Also give mentor and staff roles access in case they need it.


mataslauzadis (approved):
Looks good!

Issue #382, opened 2021-03-25
Make profile teamStatus field a validated field


heesooy:
Since we rely on `teamStatus` to be `NOT_LOOKING` to filter out profiles in the `/search/` endpoint, we should actually validate the field during profile creation/update and use enums or constants to avoid "magic strings" in the code

PR #383, opened 2021-03-25
Change teamStatus fields to UPPER_SNAKE_CASE and fix a broken test

heesooy:
- `teamStatus` should be one of `LOOKING_FOR_MEMBERS` `LOOKING_FOR_TEAM` `NOT_LOOKING`
- Changed relevant tests and docs. One of the tests was not changed in #372  where a profile with `NOT_LOOKING` should have been excluded from the search results. The issue was that the `teamStatus` value used in the test was not the one we are actually excluding for, so it was a bad test that wasn't updated.

TODO
- #382

### DIFF
--- a/documentation/docs/reference/services/Profile.md
+++ b/documentation/docs/reference/services/Profile.md
@@ -6,6 +6,8 @@ GET /profile/
 
 Returns the profile stored for the current user.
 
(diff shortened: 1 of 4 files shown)


mataslauzadis (approved):
Looks good, I like the naming convention, more like an enum now. Also nice catch on the test case, I did not realize that "Found Team" and "Not Looking" could both just become "NOT_LOOKING".

PR #385, opened 2021-03-29
Fix incorrect routing of profile search/filtered endpoints

heesooy:
This should have been a part of #371

My bad for not catching it in CR. E2e testing would have caught this 😿 

Let's really make sure to test these by hand before pushing.


mataslauzadis (dismissed):
Also my bad -- I think I coded this before I had been blessed by Postman so I was only testing the internal services rather than the API endpoints themselves.

Maybe we can get into e2e testing after the hackathon?

heesooy:
> Also my bad -- I think I coded this before I had been blessed by Postman so I was only testing the internal services rather than the API endpoints themselves.
> 
> Maybe we can get into e2e testing after the hackathon?

Yeah, let's do it

PR #389, opened 2021-04-01
Upload autoaccept.py

mataslauzadis:
This will be used to automatically accept users as they register.
Requires an admin JWT token stored in environment variables under the name "HackIllinois_Admin_JWT"


heesooy (approved):
Thanks for getting this done so quickly!

PR #393, opened 2021-04-01
Fix/checkin logic bugs

heesooy:
- Working on merging the profile id change with the event checkin work that Dev did in #378 
- Cleaned up some redundant fields and variable names in the /event/code/ endpoints
- Moved all the db collection creation and deletion code in profile to CreateProfile and DeleteProfile.
- Disabled DELETE profile for now, as there may be some issues associated with having this endpoint in the first place (see #394)

### DIFF
--- a/documentation/docs/reference/services/Event.md
+++ b/documentation/docs/reference/services/Event.md
@@ -386,7 +386,6 @@ Updates a struct that contains information about the event code (generated upon 
 Request format:
 ```
 {
-    "id": "52fdfc072182654f163f5f0f9a621d72",
     "code": "new_code",
     "expiration": 1521388800
 }
--- a/documentation/docs/reference/services/Event.md
+++ b/documentation/docs/reference/services/Event.md
@@ -362,8 +362,6 @@ Response format:
 }
 ```
 
-
-
 GET /event/code/{id}/
 ----------------------------
 
@@ -404,12 +402,13 @@ Response format:
 }
 ```
 
-
 POST /event/checkin/
 ----------------------------
 
 Retrieves a struct that contains information about the event checkin status, point increment value, and total point number.
 Takes in a struct that contains an event checkin code.
+
+Valid values for `status` are `Success`, `InvalidTime`, `AlreadyCheckedIn`. When `status != Success`, the `newPoints` and `totalPoints` fields will be -1 and should be ignored.
 
 Request format:
 ```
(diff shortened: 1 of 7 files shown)


heesooy:
WIP

documentation/docs/reference/services/Event.md
```diff
@@ -404,13 +401,14 @@ Response format:
...
 ```
 
-
 POST /event/checkin/
 ----------------------------
 
 Retrieves a struct that contains information about the event checkin status, point increment value, and total point number.
 Takes in a struct that contains an event checkin code.
 
+Valid values for `status` are `Success`, `InvalidTime`, `AlreadyCheckedIn`. When `status != Success`, the `newPoints` and `totalPoints` fields will be -1 and should be ignored.
```
mataslauzadis, line 410: Is there anything stopping us from removing newPoints and totalPoints from the struct? I don't think any other endpoint uses them.

services/profile/controller/controller.go
```diff
@@ -239,13 +240,21 @@ func GetValidFilteredProfiles(w http.ResponseWriter, r *http.Request) {
 }
 
 /*
-	RedeemEvent checks the appropriate table to check whether the given event id has already been redeemed. If the event is not in the table, it add to the array.
+	RedeemEvent checks the appropriate table to check whether the given event id has already
+	been redeemed by the specified user. If the event is not in the table, it add to the array.
```
mataslauzadis, line 244: nitpick typo "add it" instead of "it add"

mataslauzadis (commented):
Looks good to me, is the DeleteProfile endpoint going to be disabled for the whole event?

mataslauzadis (approved):

documentation/docs/reference/services/Event.md
DevSatpathy: I thought it might be helpful for the frontend for updating information, but if they're already making calls to the event, it isn't as necessary (like re-rendering the profile and showing an updated point amount).

DevSatpathy (dismissed):
Haven't had a chance to checkout yet but the code lgtm

documentation/docs/reference/services/Event.md
heesooy: > Looks good to me, is the DeleteProfile endpoint going to be disabled for the whole event?

Yeah, I think it might be risky to have it available. I don't know if any front ends are using it at all. @PatrickKan

DevSatpathy (approved):

PR #395, opened 2021-04-02
Use status field instead of http error to indicate incorrect event checkin code

heesooy:
Before, we were returning an http 500 with a raw error message string indicating that the provided code was invalid (nonexistent code). We should instead return a response and populate the `status` field of the response so that front-end can more gracefully handle this expected case.


mataslauzadis (approved):
Looks good!

Issue #234, opened 2019-09-14
Add endpoint to patch user registration


benpankow:
We currently have a `PUT /registration/attendee/` endpoint that allows us to completely replace a user's registration data. If instead we want to update just one part of that data, or add a single new field, it would be good to have a `PATCH /registration/attendee/` endpoint which would let us do so.

If the initial registration looks like:
```
{
    "firstName": "John",
    "lastName": "Smith",
    "email": "john@gmail.com",
    "shirtSize": "M",
    ...
}
```

and we call `PATCH /registration/attendee/` with
```
{
    "email": "john@illinois.edu",
    ...
}
```

the final registration should look like

```
{
    "firstName": "John",
    "lastName": "Smith",
    "email": "john@illinois.edu",
    "shirtSize": "M",
    ...
}
```

PR #407, opened 2021-09-07
Feat/patch registration attendee

kgukevin:
Implements patch function for attendee registration. Adds parameter to validate to allow override of form field requirements. Removes default value population from model conversions.

### DIFF
--- a/common/datastore/conversions.go
+++ b/common/datastore/conversions.go
@@ -59,9 +59,10 @@ func toObject(raw_data interface{}, definition DataStoreDefinition) (interface{}
 			if err != nil {
 				return nil, NewErrInField(field.Name, err)
 			}
-		} else {
-			data[field.Name] = getDefaultValue(field.Type)
-		}
+		} 
+		// else {
+		// 	data[field.Name] = getDefaultValue(field.Type) //removed for registration/attendee patch function
+		// }
 	}
 
 	return data, nil
--- a/common/datastore/validation.go
+++ b/common/datastore/validation.go
@@ -5,8 +5,11 @@ import (
 	"gopkg.in/go-playground/validator.v9"
 )
 
-func (datastore *DataStore) Validate() error {
+func (datastore *DataStore) Validate(params ...int) error {
 	validate := validator.New()
+	if len(params) == 1{
+		return nil
+	}
 
 	return validateField(datastore.Data, datastore.Definition, validate)
 }
--- a/services/registration/controller/controller.go
+++ b/services/registration/controller/controller.go
@@ -21,6 +21,7 @@ func SetupController(route *mux.Route) {
 	router.HandleFunc("/attendee/", GetCurrentUserRegistration).Methods("GET")
 	router.HandleFunc("/attendee/", CreateCurrentUserRegistration).Methods("POST")
 	router.HandleFunc("/attendee/", UpdateCurrentUserRegistration).Methods("PUT")
+	router.HandleFunc("/attendee/", PatchCurrentUserRegistration).Methods("PATCH")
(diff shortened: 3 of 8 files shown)


common/datastore/validation.go
```diff
@@ -5,8 +5,11 @@ import (
 	"gopkg.in/go-playground/validator.v9"
 )
 
-func (datastore *DataStore) Validate() error {
+func (datastore *DataStore) Validate(params ...int) error {
 	validate := validator.New()
+	if len(params) == 1{
+		return nil
+	}
```
heesooy, line 12: Could you explain what this change does?

Edit: ah I see you made this flag to skip the validation. I think we can avoid having to do this since we might need to change our strategy for this PR. See my other comment below

common/datastore/conversions.go
```diff
@@ -59,9 +59,10 @@ func toObject(raw_data interface{}, definition DataStoreDefinition) (interface{}
 			if err != nil {
 				return nil, NewErrInField(field.Name, err)
 			}
-		} else {
-			data[field.Name] = getDefaultValue(field.Type)
-		}
+		} 
+		// else {
+		// 	data[field.Name] = getDefaultValue(field.Type) //removed for registration/attendee patch function
+		// }
```
heesooy, line 65: Hm. Now that I think about it, I think we actually don't want to do this. Users currently expect the API to always return all fields, even if their value is empty/default. 1. We wouldn't want to introduce a change to this behavior without thinking about its implications on the user and 2. I think the API having a consistent return format is better.

So as for how to get around this issue: we can try to manually strip off fields from the user_registration when we pass it into the database patch function

services/registration/controller/controller.go
```diff
@@ -257,6 +258,71 @@ func UpdateCurrentUserRegistration(w http.ResponseWriter, r *http.Request) {
...
+	original_registration, err := service.GetUserRegistration(id)
+
+	if err != nil {
+		errors.WriteError(w, r, errors.DatabaseError(err.Error(), "Could not get user's original registration."))
+		return
+	}
+
+	user_registration.Data["github"] = user_info.Username
+
+	user_registration.Data["createdAt"] = original_registration.Data["createdAt"]
```
heesooy, line 298: We shouldn't need to re-insert the "github" or "createdAt" fields. We only need to change the "updatedAt" field + the fields the user provided in the request.

heesooy (changes requested):
Good start! I think we might need to change our strategy here a bit, so gonna nack it but let me know if you disagree/have a different approach. Sorry for the back and forth. 

Some general tips regarding PRs:
1. I noticed your second commit `0cc6c76` has the same commit message as the first commit. I would either squash all the commits into one commit or make sure that your second commit has a different, useful message
2. No need to change the PR titles to make it perfect. As you can see every edit leaves an edit history in the PR and it makes the PR a little messy, and as long as the title makes general sense it's fine
3. If the PR addresses/fixes an [issue](https://github.com/HackIllinois/api/issues) then you can reference that issue number from anywhere in this PR, whether it's the original PR message or any comments. For example, this PR is related to #234

kgukevin:
Per offline discussion with @heesooy, the current strategy for the patch endpoint implementation was to use the raw request body as a map and compare it to the registration definition to prune unwanted excess fields before patching it to the database. This prevents default value population in the patch datastore to combat "erasing" unpopulated fields after patching while preserving the functionality for better consistency. ValidateNonEmpty was also created to accommodate for validating a patch (which is a subset of fields in the registration definition). Other strategies were discussed to combat default values (such as getting original registration from the database and updating it with the new data before patching back to the database) but the implemented strategy was preferred for database operation atomicity.

PR #417, opened 2021-10-26
Refactor misleading database errors

tanyongzhi:
This PR fixes misleading database errors. In particular, this PR, when applied, will:

**- Fix mongo database initialization hanging during local development if local mongo instance is not started**
  - In development, the timeout for database connection is 0 - corresponding to a blocking call to DialWithInfo(). This effectively leads to the service being unresponsive, as the services will forever be waiting on the connection to the database during initialization without logging anything meaningful. As a result, all subsequent requests to the service will simply be timed out and return 500 errors. 
  - This is fixed by setting the timeout for database connections to 60 both in production and in development. Correspondingly, DialWithInfo() will return an error, which will be logged. In the future, we can consider exiting the service upon initialization failure.

**- Change not found errors to return 404 instead of 500**
 - Previously, many GET requests that retrieve data from mongo will return a 500 error if the data is not present in the database. This is misleading, and 500 errors should be reserved for internal connection errors with the database backend. This is fixed by changing the database error to a 404 status code if the database returns a not found error.

### DIFF
--- a/common/errors/DatabaseError.go
+++ b/common/errors/DatabaseError.go
@@ -1,7 +1,6 @@
 package errors
 
 import (
-	"fmt"
 	"net/http"
 )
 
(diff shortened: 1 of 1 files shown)


TakshPSingh:
@tanyongzhi do you think it might be a better idea to move the 404 error part into a separate commit so that we can merge it now while the other part of this commit is being reviewed?

PR #423, opened 2021-10-29
Fix decision/filter/ unable to filter by wave

tanyongzhi:
Previously, POST /decision/filter/ by wave did not work, and the entire decision list was returned regardless of the filter value. This is because the wave field in the request model is encoded as an int, and there is only database query filter generation logic (in filter.go.CreateFilterQuery) for int64. 

This PR fixes this by adding a case for int data types in the filtering logic. It is handled exactly like how int64s data types are handled.

### DIFF
--- a/common/database/filter.go
+++ b/common/database/filter.go
@@ -149,7 +149,7 @@ func CreateFilterQuery(
 		case "string":
 			qs, err = UpdateQuerySelectorString(qs, query_type, values)
 			query[key] = qs
-		case "int64":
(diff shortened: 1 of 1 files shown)


TakshPSingh (approved):
LGTM! Ideally, we'd just use int64 everywhere for consistency, but this seems like the cleanest way to do this unless we were to refactor the entire codebase (which would be a lot of work).

DevSatpathy (approved):

PR #425, opened 2021-11-02
Updated docs for DELETE /profile/

Nydauron:
Added "Temporarily disabled" to top of description for DELETE /profile/.


mataslauzadis (approved):
We should discuss if we should permanently remove this endpoint or not.

DevSatpathy (approved):

PR #427, opened 2021-11-04
Add conflict errors for duplicate registrations

TakshPSingh:
This PR fixes the registration service to return 409 instead of 500 when it encounters a duplicate registration.


services/registration/service/registration_service.go
```diff
@@ -14,6 +14,8 @@ var validate *validator.Validate
 
 var db database.Database
 
+const RegistrationAlreadyExists = "Registration already exists."
```
TakshPSingh, line 17: I don't the love that fact that I'm using strings to pass around errors. It might be a better idea to create a separate class for errors like this in the common package. Thoughts?

services/registration/controller/controller.go
```diff
@@ -126,6 +126,10 @@ func CreateCurrentUserRegistration(w http.ResponseWriter, r *http.Request) {
 	err = service.CreateUserRegistration(id, user_registration)
 
 	if err != nil {
+		if err.Error() == service.RegistrationAlreadyExists {
+			errors.WriteError(w, r, errors.ConflictError(err.Error(), "Could not create user registration."))
```
TakshPSingh, line 130: I apologize if the control flow is a little messy. I'll fix it after the DB level fix for 404 errors is merged (#417)

services/registration/service/registration_service.go
tanyongzhi: I think this is fine. The error types in the common dir seem to be for controller level errors, so I think it's best to keep things consistent (ie. using strings for service level errors, and err types for controller level errors)

TakshPSingh: Yeah, that makes sense.

PR #429, opened 2021-11-11
Initial Integration Test

mataslauzadis:
Add 10 users, accept them, finalize them


DevSatpathy (approved):

Makefile
```diff
@@ -102,3 +102,12 @@ container-push:
 .PHONY: docs
 docs:
 	$(MAKE) -C $(REPO_ROOT)/documentation build
+
+.PHONY: test-e2e
+test-e2e:
+	@echo "Beginning end-to-end tests";
+	@echo "Checking if the API is running...";
+	@curl --silent --output /dev/null localhost:8000 || (echo "Failed to connect to the API. Is it running?"; exit 1;)
```
tanyongzhi, line 110: Can we start the api here? Similar to how we're doing it for `make test`

tanyongzhi:
In my opinion, we should add some database clean up methods after the test ends. Right now, the data in the databases is not cleared after the test ends, so different runs of the tests aren't independent of each other (eg. the second run of this test will lead to a "Cannot modify finalized decisions" error)

Makefile
```diff
@@ -102,3 +102,12 @@ container-push:
...
+	@echo "Running end-to-end tests";
+	@HI_CONFIG=file://$(REPO_ROOT)/config/test_config.json go test $(BASE_PACKAGE)/services/user/tests/end_to_end || exit 1;
```
tanyongzhi: Since the api is started before the `HI_CONFIG` environment variable is set to test config here, the api would still be using the dev config

mataslauzadis:
> In my opinion, we should add some database clean up methods after the test ends. Right now, the data in the databases is not cleared after the test ends, so different runs of the tests aren't independent of each other (eg. the second run of this test will lead to a "Cannot modify finalized decisions" error)

I agree, we talked about two possible solutions in the meeting:
1) Use test config which will put all of the new data in the test database (does this get wiped after the test is over?)
2) Start a transaction to MongoDB at the beginning of the test and revert it at the end

tanyongzhi:
> > In my opinion, we should add some database clean up methods after the test ends. Right now, the data in the databases is not cleared after the test ends, so different runs of the tests aren't independent of each other (eg. the second run of this test will lead to a "Cannot modify finalized decisions" error)
> 
> I agree, we talked about two possible solutions in the meeting:
> 
> 1. Use test config which will put all of the new data in the test database (does this get wiped after the test is over?)
> 2. Start a transaction to MongoDB at the beginning of the test and revert it at the end

Yeah, I think the test database approach is probably the best. We can wipe the data after the test is completed to ensure a clean slate for every run of the test.
I'm not too sure how Mongo transactions work, but it seems that we have to change the existing database code in the service to use transactions (more info [here](https://www.mongodb.com/developer/quickstart/golang-multi-document-acid-transactions/)).

Makefile
mataslauzadis: I solved this by adding a new Make target called `run-test` which launches the API using the test config. Then, when the `integration-test` target is made, it has `run-test` as a dependency. But there is an issue I have been unable to solve involving a race condition, sometimes the API does not fully start before the `integration-test` target tries to run its tests, so it fails. 

I think these should be separate issues since the integration tests are now working, you just have to launch the API yourself on the side.

tests/add_accept_finalize_user_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	return session
+}
+
+func TestMain(m *testing.M) {
+	admin_client = GetAdminClient()
+
+	mongo_session = GetLocalMongoSession()
+	mongo_session.DB("user").DropDatabase()
+	mongo_session.DB("decision").DropDatabase()
```
tanyongzhi, line 57: Tests are still failing on subsequent runs after the first run.

I think that this cleanup is not being handled properly. We should be using the test db names in the config file (test-user, test-decision, etc.) instead of the dev db names. We can load the test config in `TestMain` and use `config.USER_DB_NAME` and `config.DECISION_DB_NAME` to specify the correct db names.

mataslauzadis: The API is still being run using the development config (`make run`) since I did not switch to using the test config in this PR. So the correct databases are being dropped here. But I agree we should use the config instead.

tanyongzhi: Ah, I see. I assumed that the API is being started by `scripts/run-test.sh `. But yes, we can change this in another PR.

tanyongzhi (approved):
LGTM. Future PR to change database being dropped to the test database.

mataslauzadis:
I made the changes to drop the databases using the names from the config. Now, if you follow these steps it should work.

1. Start the API using `make run-test`
2. Run `make integration-test`

A future PR will remove the need to start the API (step 1) by making the `integration-test` target start and kill the API itself

tanyongzhi (approved):

DevSatpathy (approved):

PR #430, opened 2021-12-24
Event point system

Nydauron:
- Event model now have 3 fields for points: `InPersonPoints`, `InPersonVirtPoints`, and `VirtualPoints`
- Event code model has an additional field `IsVirtual` to distinguish between code for virtual/in-person use
- Changed `PUT /event/code/{id}` to `POST /event/code/`
- Added an admin-only internal route `GET /profile/user/{id}/` so that we can determine if the user redeeming the event is virtual or in-person
- Fixed a bug where new entries into `profileattendance` were resulting in duplicate rows with empty `id` fields
- Reimplemented `POST /event/checkin` to check the user for virtual status
- Updated tests for profile and event service
- Updated docs for profile and event

### DIFF
--- a/documentation/docs/reference/services/Profile.md
+++ b/documentation/docs/reference/services/Profile.md
@@ -6,8 +6,6 @@ GET /profile/
 
 Returns the profile stored for the current user.
 
-Valid values for ``teamStatus`` are ``LOOKING_FOR_MEMBERS``, ``LOOKING_FOR_TEAM``, and ``NOT_LOOKING``.
-
 ***The `id` in the profile service refers to a separate, randomly-generated, profile-only id. This is different from the (user) `id` used in other services. When a profile is created, a mapping from the user `id` to the profile `id` is stored in the database.***
 
 We will distinguish user id and profile id by using `github123456` and `profileid123456` for each, respectively, in the examples below.
@@ -19,12 +17,7 @@ Response format:
     "firstName": "John",
     "lastName": "Doe",
     "points": 2021,
-    "timezone": "Americas UTC+8",
-    "avatarUrl": "https://github.com/.../profile.jpg",
-    "discord": "patrick#1234",
-    "teamStatus": "LOOKING_FOR_TEAM",
-    "description": "Lorem Ipsum…",
-    "interests": ["C++", "Machine Learning"]
+    "isVirtual": false
 }
 
 ```
@@ -32,7 +25,7 @@ Response format:
 GET /profile/{id}/
 -------------------------
 
-Returns the profile stored for user that has the ID ``{id}``.
+Returns the profile stored for user that has the profile ID ``{id}``.
 
 Response format:
 ```
@@ -41,16 +34,29 @@ Response format:
     "firstName": "John",
     "lastName": "Doe",
     "points": 2021,
-    "timezone": "Americas UTC+8",
-    "avatarUrl": "https://github.com/.../profile.jpg",
-    "discord": "patrick#1234",
-    "teamStatus": "LOOKING_FOR_TEAM",
-    "description": "Lorem Ipsum…",
-    "interests": ["C++", "Machine Learning"]
+    "isVirtual": false
 }
 ```
 
-GET /profile/list/?teamStatus=value&interests=value,value,value&limit=value
+GET /profile/user/{id}/
+-------------------------
+
+**Internal use only.**
+
+Returns the profile stored for user that has the user ID ``{id}``.
+
+Response format:
+```
+{
+    "id": "profileid123456",
+    "firstName": "John",
+    "lastName": "Doe",
+    "points": 2021,
+    "isVirtual": false
+}
+```
+
+GET /profile/list/?isVirtual=value&interests=value,value,value&limit=value
 -------------------------
 
 **Internal use only.**
@@ -60,8 +66,6 @@ Returns a list of profiles matching the filter conditions.
 teamStatus is a string matching the user's team status.
 
 interests is a comma-separated string representing the user's interests.
-
-- i.e if the user's interests are ["C++", "Machine Learning"], you can filter on this by sending ``interests="C++,Machine Learning"``
 
 If a ``limit`` parameter is provided, it will return the first matching ``limit`` profiles. Otherwise, it will return all of the matched profiles.
 
@@ -76,24 +80,14 @@ Response format:
             "firstName": "John",
             "lastName": "Doe",
             "points": 2021,
-            "timezone": "Americas UTC+8",
-            "avatarUrl": "https://github.com/.../profile.jpg",
-            "discord": "patrick#1234",
-            "teamStatus": "LOOKING_FOR_TEAM",
-            "description": "Lorem Ipsum…",
-            "interests": ["C++", "Machine Learning"]
+            "isVirtual": false
         },
         {
             "id": "profileid123456",
             "firstName": "John",
             "lastName": "Doe",
             "points": 2021,
-            "timezone": "Americas UTC+8",
-            "avatarUrl": "https://github.com/.../profile.jpg",
-            "discord": "patrick#1234",
-            "teamStatus": "LOOKING_FOR_MEMBERS",
-            "description": "Lorem Ipsum…",
-            "interests": ["C++", "Machine Learning"]
+            "isVirtual": false
         },
     ]
 }
@@ -109,12 +103,7 @@ Request format:
 {
     "firstName": "John",
     "lastName": "Doe",
-    "timezone": "Americas UTC+8",
-    "avatarUrl": "https://github.com/.../profile.jpg",
-    "discord": "patrick#1234",
-    "teamStatus": "LOOKING_FOR_TEAM",
-    "description": "Lorem Ipsum…",
-    "interests": ["C++", "Machine Learning"]
+    "isVirtual": false
 }
 ```
 
@@ -124,13 +113,8 @@ Response format:
     "id": "profileid123456",
     "firstName": "John",
     "lastName": "Doe",
-    "points": 2021,
-    "timezone": "Americas UTC+8",
-    "avatarUrl": "https://github.com/.../profile.jpg",
-    "discord": "patrick#1234",
-    "teamStatus": "LOOKING_FOR_TEAM",
-    "description": "Lorem Ipsum…",
-    "interests": ["C++", "Machine Learning"]
+    "points": 0,
+    "isVirtual": false
 }
 ```
 
@@ -144,14 +128,10 @@ Note you can not edit the ``points`` field through this.
 Request format:
 ```
 {
+    "id": "profileid123456",
     "firstName": "John",
     "lastName": "Doe",
-    "timezone": "Americas UTC+8",
-    "avatarUrl": "https://github.com/.../profile.jpg",
-    "discord": "patrick#1234",
-    "teamStatus": "LOOKING_FOR_TEAM",
-    "description": "Lorem Ipsum…",
-    "interests": ["C++", "Machine Learning"]
+    "isVirtual": true
 }
 ```
 
@@ -161,13 +141,8 @@ Response format:
     "id": "profileid123456",
     "firstName": "John",
     "lastName": "Doe",
-    "points": 2021,
-    "timezone": "Americas UTC+8",
-    "avatarUrl": "https://github.com/.../profile.jpg",
-    "discord": "patrick#1234",
-    "teamStatus": "LOOKING_FOR_TEAM",
-    "description": "Lorem Ipsum…",
-    "interests": ["C++", "Machine Learning"]
+    "points": 0,
+    "isVirtual": true
 }
 ```
 
@@ -186,13 +161,8 @@ Response format:
     "id": "profileid123456",
     "firstName": "John",
     "lastName": "Doe",
-    "points": 2021,
-    "timezone": "Americas UTC+8",
-    "avatarUrl": "https://github.com/.../profile.jpg",
-    "discord": "patrick#1234",
-    "teamStatus": "LOOKING_FOR_TEAM",
-    "description": "Lorem Ipsum…",
-    "interests": ["C++", "Machine Learning"]
+    "points": 0,
+    "isVirtual": false
 }
 ```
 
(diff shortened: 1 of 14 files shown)


services/event/service/event_service.go
```diff
@@ -487,42 +487,44 @@ func CanRedeemPoints(event_code string) (bool, string, error) {
...
 
-	return current_time < expiration_time, eventCode.ID, nil
+	return current_time < expiration_time, eventCode.IsVirtual, eventCode.ID, nil
 }
 
 /*
 	Returns the eventcode struct for the event with the given id
 */
-func GetEventCode(id string) (*models.EventCode, error) {
+func GetEventCode(id string) (*[]models.EventCode, error) {
```
mataslauzadis, line 502: we should change the name of this function to `GetEventCodes` since it returns a list now

services/event/service/event_service.go
```diff
@@ -487,42 +487,44 @@ func CanRedeemPoints(event_code string) (bool, string, error) {
...
 		return nil, err
 	}
 
-	return &eventCode, nil
+	return &eventCodes, nil
 }
 
 /*
 	Updates the event code and end time with the given id
+	If no matching code is not found, then it is a new code and add it
```
mataslauzadis, line 520: should it be "If no matching code is found"?

services/event/service/event_service.go
```diff
@@ -487,42 +487,44 @@ func CanRedeemPoints(event_code string) (bool, string, error) {
...
 */
-func UpdateEventCode(id string, eventCode models.EventCode) error {
+func UpdateEventCode(code string, eventCode models.EventCode) error {
 	selector := database.QuerySelector{
-		"id": id,
+		"code": code,
```
mataslauzadis, line 524: I am wondering why the selector was changed from `id` to `code`?

correct me if I'm wrong, but since we are trying to update the `code` of the event, and you are also searching the database for an event with the new `code` instead of the old `code` (or ideally, the unique `id`), it won't find the correct event.

The old code worked because we grabbed the event ID from PUT /event/code/**{id}** but since you removed that, I feel we are missing some information that is necessary to do this database update

services/event/service/profile_service.go
```diff
@@ -53,3 +53,19 @@ func AwardPoints(id string, points int) (*models.Profile, error) {
 
 	return &profile, nil
 }
+
+func GetIsUserVirtual(id string) (bool, error) {
+	var profile models.Profile
+
+	status, err := apirequest.Get(config.PROFILE_SERVICE+"/profile/user/"+id+"/", &profile)
```
mataslauzadis, line 60: you may be able to function call `GetProfile` directly here instead of making a GET request to ourselves

services/profile/controller/controller.go
```diff
@@ -72,6 +74,29 @@ func GetProfileById(w http.ResponseWriter, r *http.Request) {
...
+/*
+	GetProfileByUserId is used to get a profile for a provided user id.
+*/
+func GetProfileByUserId(w http.ResponseWriter, r *http.Request) {
+	user_id := mux.Vars(r)["id"]
+
+	profile_id, err := service.GetProfileIdFromUserId(user_id)
+
+	if err != nil {
+		errors.WriteError(w, r, errors.DatabaseError(err.Error(), "Could not get profile id associated with the user"))
```
mataslauzadis, line 86: should we add "associated with the user **{user_id}**"? it gives better error information

services/profile/service/profile_service.go
```diff
@@ -295,9 +295,10 @@ func GetFilteredProfiles(parameters map[string][]string) (*models.ProfileList, e
 /*
 	Returns a list of profiles filtered upon teamStatus and interests. Will be limited to only include the first "limit" results.
 	Will also remove profiles with a TeamStatus set to "NOT_LOOKING"
+
+	Sorta redundant since the Profile model doesn't have team status or interest fields anymore
```
mataslauzadis, line 299: I agree, we can make an Issue to clean this up.

mataslauzadis (commented):
Looks very good. I had a few comments.

services/event/service/event_service.go
Nydauron: Ok so, I understand the issue at hand, but with the current setup, the table will have multiple fields where `id` can be equal (since events can have a virtual and an in-person code). The idea behind switching the primary key to `code` is to fix this unique identifier issue.

What I'm realizing now is that while the current `POST` route works perfectly for updating information for `id`, `isVirtual`, and `expiration`, we can't change `code` without deleting the old one and creating a new one. A potential solution could be to make the route `POST /event/code/{old_code_id}`. That way we can uniquely identify which code to modify and have the ability to modify the code id.

Also, for clarity in the model, I'll probably change the `EventCode` struct to:
```golang
type EventCode struct {
	CodeID     string `json:"codeID"`
	EventID    string `json:"eventID"`
	IsVirtual  bool   `json:"isVirtual"`
	Expiration int64  `json:"expiration"`
}
```

services/event/service/profile_service.go
Nydauron: `GetProfile` is not part of the Event service, so my thought was we cannot (and probably should not) make calls to functions across services since these are separate processes with separate databases. Hence, why I used a request.

There are some other instances of service methods within the code base where requests are used to retrieve information across services.

mataslauzadis: Yep you're right, I was looking in the wrong service

services/event/service/event_service.go
mataslauzadis: Yeah I think using {old_code} is a great way to fix it. I think we have two options.

1. Change the endpoint to POST /event/code/**{old_code_id}** so we can update the correct object
2. Don't allow API users to modify the code at all

I think 2 is a feasible option because once the random code is generated once, I don't see many reasons to change it afterwards.

Nydauron: Alright then. In both cases, would it be good to also have a DELETE route? Right now, the only way a code gets deleted is if its expiration datetime exceeds the current time and some user tries to utilize said code.

mataslauzadis: Yes that would probably be good to have, but not necessary for this PR in my opinion

services/event/models/event.go
```diff
@@ -1,15 +1,17 @@
...
+	Name               string          `json:"name"                validate:"required"`
+	Description        string          `json:"description"         validate:"required"`
+	StartTime          int64           `json:"startTime"           validate:"required"`
+	EndTime            int64           `json:"endTime"             validate:"required"`
+	Locations          []EventLocation `json:"locations"           validate:"required,dive,required"`
+	Sponsor            string          `json:"sponsor"`
+	EventType          string          `json:"eventType"           validate:"required,oneof=MEAL SPEAKER WORKSHOP MINIEVENT QNA OTHER"`
+	InPersonPoints     int             `json:"inPersonPoints"`
+	InPersonVirtPoints int             `json:"inPersonVirtPoints"`
+	VirtualPoints      int             `json:"virtualPoints"`
```
tanyongzhi, line 14: Nit: make all (or one, and add defaults) of these a required field too?

services/event/controller/controller.go
```diff
@@ -162,47 +162,38 @@ func GetEventCode(w http.ResponseWriter, r *http.Request) {
...
 		errors.WriteError(w, r, errors.DatabaseError(err.Error(), "Failed to receive event code information from database"))
 		return
 	}
 
-	json.NewEncoder(w).Encode(code)
+	json.NewEncoder(w).Encode(codes)
 }
 
 /*
 	Endpoint to update an event code and end time
```
tanyongzhi, line 176: Nit: change "update" to "upsert"

services/event/controller/controller.go
```diff
@@ -26,7 +26,7 @@ func SetupController(route *mux.Route) {
 	router.HandleFunc("/", UpdateEvent).Methods("PUT")
 	router.HandleFunc("/", GetAllEvents).Methods("GET")
 	router.HandleFunc("/code/{id}/", GetEventCode).Methods("GET")
```
tanyongzhi, line 28: Nit: change to "GetEventCodes", similar to the service level function

services/event/service/event_service.go
tanyongzhi: I agree with Matas on not letting users to modify the code id. We should state this explicitly though, in an explicit error message, and a note in the documentation.

tanyongzhi (dismissed):
LGTM, left some nits

services/event/models/event.go
Nydauron: It appears that the `required` tag checks if the parameter is not zero if the data type is a number. Also, I'm not entirely too sure if there is a value that will always be set. We could have events that are in-person exclusive, virtual-only, or a combination of the two, so not all point values are guaranteed will have a value greater than 0.

services/event/service/event_service.go
Nydauron: Alrighty, went ahead and implemented option 2. Also added a little note `POST /event/code` in the docs saying that codes can't be updated.

Nydauron: > We should state this explicitly though, in an explicit error message, and a note in the documentation.

As of right now, it's impossible to determine if the user is attempting to change `code`. Since the method is essentially upserting, if the code simply doesn't exist, it creates a new row.

A possible solution might be to do something like `POST /event/code/{code}`, but I feel then it would be a bit counterintuitive to not implement option 1.

services/event/models/event.go
Nydauron: Update: we could do something like:
```suggestion
        InPersonPoints     int             `json:"inPersonPoints"      validate:"required_without=VirtualPoints"`
	InPersonVirtPoints int             `json:"inPersonVirtPoints"  validate:"required_with_all=InPersonPoints VirtualPoints"`
	VirtualPoints      int             `json:"virtualPoints"       validate:"required_without=InPersonPoints"`
}
```

Only issue that validator.v9 is giving some issues with it:
```bash
2022/01/03 03:04:18 http: panic serving 127.0.0.1:48154: Undefined validation function 'required_without' on field 'InPersonPoints'
```

Fortunately, validator.v10 worked for me.

services/event/service/event_service.go
mataslauzadis: I think it is possible to prevent the user from changing the event struct's `code` here.

After creating the selector, you can use `findOne` to get the old event, then set the new `eventCode` struct's `code` field to that, and then finally do the upsert.

Here is a rough idea of what I was thinking
```go
selector := database.QuerySelector{
	"codeid": code,
}
// load the old code
var oldEventCode models.EventCode  
err := db.FindOne("eventcodes", selector, &oldEventCode)

// update the code field to make sure it is not changed by the user
eventCode.code = oldEventCode.code

// then do the upsert as normal
_, err := db.Upsert("eventcodes", selector, &eventCode)
```

tanyongzhi: > ```go
> selector := database.QuerySelector{
> 	"codeid": code,
> }
> ```

Is this assuming that the `code` variable is the old code? That would be implementing option 1), right?

In my opinion, I think it is alright to create a new row if a new code is specified. Putting the responsibility of calling this endpoint with the correct code can be placed on the caller, instead.

services/event/models/event.go
tanyongzhi: Sounds good. To clarify, are we assuming an event with no `InPersonPoints` or `VirtualPoints` will correspond to 0 points being awarded for participating in the event in person and virtually respectively?

services/event/service/event_service.go
Nydauron: > I think it is possible to prevent the user from changing the event struct's `code` here.
> 
> After creating the selector, you can use `findOne` to get the old event, then set the new `eventCode` struct's `code` field to that, and then finally do the upsert.
> 
> Here is a rough idea of what I was thinking
> 
> ```go
> selector := database.QuerySelector{
> 	"codeid": code,
> }
> // load the old code
> var oldEventCode models.EventCode  
> err := db.FindOne("eventcodes", selector, &oldEventCode)
> 
> // update the code field to make sure it is not changed by the user
> eventCode.code = oldEventCode.code
> 
> // then do the upsert as normal
> _, err := db.Upsert("eventcodes", selector, &eventCode)
> ```
> 
> You can wrap the `findOne` in an if-statement. If the eventCode does not already exist, you can run the upsert directly

The current implementation works perfectly fine. In all cases when something is being upserted, it will either be updated if `code` is found or inserted otherwise.

For clarification on what I said, I was building on what Yong commented on; if we were to return some error to the user saying that they are modifying the code, we would probably need something like `POST /event/code/{old_code}`. It's rather a small nit really since it should be clear that the code can't be changed.

services/event/models/event.go
Nydauron: Yes, that's what will happen. Something to note is that some events are not meant to have points awarded to users since the event could be representing food/meals for example.

So, for right now, I'll just leave it as is, but if we update to validator/v10, the following tags should be good:
```golang
	InPerson     uint            `json:"inPerson"`
	InPersonVirt uint            `json:"inPersonVirt"  validate:"required_with_all=InPerson Virtual"`
	Virtual      uint            `json:"virtual"`
```

tanyongzhi: Yup, makes sense. Having 0 as the default value is reasonable, and shouldn't be too disastrous if the caller forgets to set certain point values.

tanyongzhi (approved):

documentation/docs/reference/services/Profile.md
```diff
@@ -19,20 +17,15 @@ Response format:
     "firstName": "John",
     "lastName": "Doe",
     "points": 2021,
-    "timezone": "Americas UTC+8",
-    "avatarUrl": "https://github.com/.../profile.jpg",
-    "discord": "patrick#1234",
-    "teamStatus": "LOOKING_FOR_TEAM",
-    "description": "Lorem Ipsum…",
-    "interests": ["C++", "Machine Learning"]
+    "isVirtual": false
```
DevSatpathy, line 20: We need to add back the avatar URL, so we can randomly generate a profile picture and affix it to the profile.

Nydauron: > we can randomly generate a profile picture and affix it to the profile

So, are we setting this to some URL on profile creation?

PR #432, opened 2021-12-30
Add registration metrics

tanyongzhi:
This commit adds metric collection functionality to the registration endpoint. This is meant to be a test before adding it to the rest of the endpoints. Specifically, this commit will:

- Change `HandleFunc` in the endpoint controller router to use `RegisterHandler`, a wrapper that passes the request through metrics collection middleware before moving to specific service level functionality
-  Add two metrics defined in `common/metrics/metrics_counter.go`
    -  `http_requests_total` collecting the number of http requests by endpoint and its result status code
    - `http_request_duration_seconds` collecting the duration of each http request

### DIFF
--- a/common/metrics/metrics.go
+++ b/common/metrics/metrics.go
@@ -1,0 +1,22 @@
+package metrics
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+func metricsWrapper(next http.HandlerFunc, endpoint string) http.Handler {
+	labels := map[string]string{
+		"endpoint": endpoint,
+	}
+	return promhttp.InstrumentHandlerDuration(Duration.MustCurryWith(labels),
+		promhttp.InstrumentHandlerCounter(TotalRequests.MustCurryWith(labels),
+			next))
+}
+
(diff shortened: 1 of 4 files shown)


mataslauzadis:
Would it be possible to use the current Alice middleware chain for metrics collection as well? Then we would not have to reformat each API route in each service controller, only at the gateway level.

Here is an [example of the current middleware chains for the Registration service](https://github.com/HackIllinois/api/blob/staging/gateway/services/registration.go)

And [here is where we have the middleware code](https://github.com/HackIllinois/api/tree/staging/gateway/middleware) (identification and authentication). I think we could add your metrics collection code here?

tanyongzhi:
> Would it be possible to use the current Alice middleware chain for metrics collection as well? Then we would not have to reformat each API route in each service controller, only at the gateway level.

Adding metrics collection to the Alice middleware chain at the gateway would be adding collecting the gateway metrics. However, we are interested in metrics collection per service (in this case, the registration service), which is separate from the gateway, which we assume can be running on a different machine from the service altogether.

In fact, the metrics collection should not touch or affect the gateway at all. Right now, we are collecting metrics directly via the `/metrics` endpoint in the service. 

Hope this explanation is clear!

DevSatpathy (approved):

PR #441, opened 2022-01-23
Add unauthorized action e2e test

tanyongzhi:
This PR adds a new e2e test that checks if users can perform unauthorized actions. This PR also refactors the existing e2e testing framework. Thus, **please merge this PR first before adding new e2e tests.**

In particular, when merged, this PR will:
- Refactor commonly used test functions to a common file
- Refactor token generation for e2e tests to not use `make setup`. The new token is generated by calling the `accountgen` and `tokengen` binaries directly. This is done for two reasons:
	- Previous `make integration-test` behavior was dependent on the dir which the command is executed from
	- Enable generating tokens with roles other than admin roles (eg. a token with attendee role to test unauthorized endpoint access)
- Refactor each e2e test to have its own `TestMain`
	- `make integration-test` will iterate through all these `TestMain`s, similar to `make test` for the unit tests for the services
- Add e2e test for attendee token performing authorized access

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,6 @@ docs:
 integration-test:
 	@echo "Beginning integration tests";
 	@echo "Checking if the API is running...";
-	@curl --silent --output /dev/null localhost:8000 || (echo "Failed to connect to the API. Is it running?"; exit 1;)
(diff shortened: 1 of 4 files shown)


DevSatpathy (approved):
LGTM, next thing to do is making sure the github actions are setup for this.

PR #443, opened 2022-01-27
Add profile tier to min point threshold endpoint

tanyongzhi:
- Add a new endpoint in the profile service, GET /profile/tier/threshold/, which will return the profile tier (represented in a string) to minimum point threshold mapping. The values are configured from the respective env config files.

### DIFF
--- a/config/dev_config.json
+++ b/config/dev_config.json
@@ -428,5 +428,16 @@
 				"fields": []
 			}
 		]
-	}
+	},
+
+	"TIER_THRESHOLDS": [
+		{
+			"name": "cookie",
+			"threshold": 0
+		},
+		{
+			"name": "cupcake",
+			"threshold": 50
+		}
+	]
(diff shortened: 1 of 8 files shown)


Nydauron:
Don't we also need to add an endpoint simply returning the name given a profile id or are we just giving this mapping for the apps to use and they do the calculations?

Nydauron (approved):
Changes look good. Just have the question above about whether there is any other endpoint needed for this feature.

tanyongzhi:
> Don't we also need to add an endpoint simply returning the name given a profile id or are we just giving this mapping for the apps to use and they do the calculations?

Yup, my understanding is that the client will perform all the necessary calculations with the mapping. The API's job is simply to provide that point threshold to tier mapping.

TakshPSingh (approved):
LGTM!

PR #444, opened 2022-02-01
Add metrics for reset of services

tanyongzhi:
This is a follow up for #432 to add metrics to the remainder of the services.


DevSatpathy (approved):
LGTM

PR #445, opened 2022-02-02
Add e2e tests for unauthenticated access

TakshPSingh:


tanyongzhi:
This seems similar to #441, is there a need to separate these two tests?

tests/e2e/unauthenticated_user_test/unathenticated_user_test.go
```diff
@@ -0,0 +1,84 @@
...
+	os.Exit(return_code)
+}
+
+func TestUnauthenticatedCalls(t *testing.T) {
+	// 1. Staff endpoint
+	received_mail_list := mail_models.MailList{}
+	response, _ := attendee_client.New().Get("/mail/list/").Receive(&received_mail_list, &received_mail_list)
+
+	if response.StatusCode != 403 {
+		t.Errorf("Unathenticated attendee able to access endpoint that requires authentication")
```
tanyongzhi, line 66: Nit: typo Unauthenticated instead of Unathenticated. Also on lines 74, 82

tests/e2e/unauthenticated_user_test/unathenticated_user_test.go
```diff
@@ -0,0 +1,84 @@
...
+func TestMain(m *testing.M) {
+
+	cfg, err := configloader.Load(os.Getenv("HI_CONFIG"))
+
+	if err != nil {
+		fmt.Printf("ERROR: %v\n", err)
+		os.Exit(1)
+	}
+
+	attendee_client = sling.New().Base("http://localhost:8000").Client(nil).Add("Authorization", "FAKE_TOKEN")
```
tanyongzhi, line 31: Can consider using the `common.GetSlingClient()` here, which generates a token with the specified role.

Example [here](https://github.com/HackIllinois/api/blob/staging/tests/e2e/unauthorized_actions_test/unauthorized_actions_test.go#L28)

TakshPSingh: Actually, I intentionally did not use it since GetSlingClient would generate a token for the client (essentially making them authenticated), which is the opposite of what we are trying to test.

tests/e2e/unauthenticated_user_test/unathenticated_user_test.go
TakshPSingh: My bad. I'll fix this.

tests/e2e/unauthenticated_user_test/unathenticated_user_test.go
tanyongzhi: Right, that makes sense

tanyongzhi (approved):
LGTM

PR #449, opened 2022-02-09
Add e2e test for event endpoints

TakshPSingh:


tests/e2e/events_test/events_test.go
```diff
@@ -0,0 +1,72 @@
...
+func TestMain(m *testing.M) {
+
+	cfg, err := configloader.Load(os.Getenv("HI_CONFIG"))
+
+	if err != nil {
+		fmt.Printf("ERROR: %v\n", err)
+		os.Exit(1)
+	}
+
+	admin_client = common.GetSlingClient("Admin")
```
tanyongzhi, line 28: Is the objective to test Admin role privileges, or Staff role privileges? For now, the privileges overlap completely, so it shouldn't be a big deal right now.

tanyongzhi (commented):
Should we test more staff endpoints besides event creation? What about the following flow:
- Create event
- Get the event code
- Update event
- Check if event is updated through GET endpoint

We can also test the event trackers, but that can be done in another PR.

tests/e2e/events_test/events_test.go
TakshPSingh: My bad. I didn't realize Admin and Staff were different roles. I'll make this change and also the flow that you suggested above.

tanyongzhi (approved):
LGTM!

PR #451, opened 2022-02-10
Fix to code duplicate redemption bug

Nydauron:
Fixed a bug where new entries into `profileattendance` were resulting in duplicate rows with empty `id` fields

### DIFF
--- a/services/profile/service/profile_service.go
+++ b/services/profile/service/profile_service.go
@@ -323,10 +323,11 @@ func RedeemEvent(profile_id string, event_id string) (*models.RedeemEventRespons
 
 	if err != nil {
 		if err == database.ErrNotFound {
-			err = db.Insert("profileattendance", &models.AttendanceTracker{
+			attended_events = models.AttendanceTracker{
 				ID:     profile_id,
 				Events: []string{},
-			})
+			}
+			err = db.Insert("profileattendance", &attended_events)
 
 			if err != nil {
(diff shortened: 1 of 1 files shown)


DevSatpathy (commented):
How does this work? Doesn't this just do the exact same behavior as the previous code?

Nydauron:
> How does this work? Doesn't this just do the exact same behavior as the previous code?

This was previously mentioned in a Slack post, but I'll repost here:

> UPDATE: at first i thought this commit was it, then i thought i just made a commit that changed nothing, then upon further inspection this DOES in fact fix the bug https://github.com/HackIllinois/api/pull/430/commits/b1223ec2bd45e70d07938f00033d6a2aceb86ba3
What was happening was that POST /event/checkin/ needs to call POST /profile/event/checkin/ in order to make sure not to double redeem an event. It checks this by trying to find an existing entry first. If none exists, it creates a new entry like below:

```
{
    id: profile_id,
    events: []
}
```
> However, it this is the end of the error checking and assumes the variable attended_events has a valid entry from profileattendance. Ovbsly, if the initial find failed, attended_events is blank. So, when it appends the new event id and updates the collection with selector `{id: profile_id}`, it replaces the newly added entry with:
```
{
    id: '',
    events: [event_id]
}
```

DevSatpathy (approved):

Issue #404, opened 2021-04-28
Persistent Mongo Connections


PatrickKan:

tanyongzhi:
Check api channel in the HackIllinois Slack for more information

PR #459, opened 2022-02-13
Add pool limit for mongodb connection

tanyongzhi:
This PR limits the number of our pooled connections to Mongo.

Testing with calling the `POST /registration/attendee/` endpoint 300 times concurrently:
- No PooLimit - 395 connections to Mongo
- PoolLimit = 50 - 154 connections to Mongo
- PooLimit = 25 - 132 connections to Mongo

Note that this test was done locally.
I chose the value of 25 for now, although suggestions for other pool limit values are welcome

### DIFF
--- a/common/database/mongo_database.go
+++ b/common/database/mongo_database.go
@@ -54,6 +54,7 @@ func (db *MongoDatabase) Connect(host string) error {
 	}
 
 	session, err := mgo.DialWithInfo(dial_info)
+	session.SetPoolLimit(25)
 
 	if err != nil {
 		return ErrConnection


PatrickKan (approved):
Thanks for the research! This problem has hindered the api for multiple years now but super cool that it's a one line fix to set a connection pool

tanyongzhi:
> Fixes #404 ? (Or does the rate limit need to be in place to truly fix it?)

I think we should leave #404 open for now, but perhaps rename it and provide a little bit more context in the issue description, since the issue with Mongo is a symptom of the much larger problem of rate limiting requests.

Let me know if you think that this is a good approach!

PatrickKan:
Sounds good to keep the issue open, can def try to stress test the API once these changes are merged

DevSatpathy (approved):

PR #460, opened 2022-02-17
Add async events

TakshPSingh:
By default, all events are synchronous unless the isAsync field is true.

If isAsync is true, startTime and endTime are optional.

This pull request also updates our validator's version, since the required_if field was not in the old version that we were using.


tanyongzhi (dismissed):

Nydauron (commented):
Looks good. Small nit, but `isAsync` should be added to other parts of the event documentation, not just for POST /event/.

TakshPSingh:
> Looks good. Small nit, but `isAsync` should be added to other parts of the event documentation, not just for POST /event/.

Good catch. I ended up moving the description of isAsync to the top of the page, so that frontend teams know that it applies to all event endpoints.

DevSatpathy (approved):

PR #464, opened 2022-02-20
Added script to generate top N and select at most M people from each tier

Nydauron:
As requested, this is a Python script to be used for Shark Tank. It pulls all profiles, sorts all the profiles into their respective tiers, and picks N (passed in as an arg) number of profiles per tier randomly.

If needed, `python3 sharktank.py -h` for help message.


tanyongzhi (approved):
nice work!

DevSatpathy (approved):

PR #495, opened 2022-10-25
Added endpoint tests for GET and POST profile

asehgal220:


tests/e2e/profile_test/profile_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	cfg, err := configloader.Load(os.Getenv("HI_CONFIG"))
+
+	if err != nil {
+		fmt.Printf("ERROR: %v\n", err)
+		os.Exit(1)
+	}
+
+	admin_client = common.GetSlingClient("Admin")
+	unauthenticated_client = sling.New().Base("http://localhost:8000").Client(nil).Add("Authorization", "FAKE_TOKEN")
```
Timothy-Gonzalez, line 30: Could we consider taking this `unauthenticated_client` out into `tests/common` so other tests can use it? Seems general enough to have a `common.GetSlingClient("Unauthenticated")`.

tests/e2e/profile_test/get_profile_test.go
```diff
@@ -0,0 +1,73 @@
...
+	endpoint_address := fmt.Sprintf("/profile/%s/", "12345")
+
+	recieved_profile := profile_models.Profile{}
+	response, err := admin_client.New().Get(endpoint_address).ReceiveSuccess(&recieved_profile)
+
+	if err != nil {
+		t.Errorf("Unable to make request")
+	}
+	if response.StatusCode != http.StatusOK {
+		t.Errorf("Request returned HTTP error %d", response.StatusCode)
```
Timothy-Gonzalez, line 34: It probably makes more sense to do `response.Status` since that will give the fully expanded HTTP error. `500 => 500 Internal Server Error`. Something to consider at least, applies to all the other `response.StatusCode`s.

Timothy-Gonzalez (approved):
LGTM, only a few minor changes I'd make but those are mainly just nitpicks and can be done later if need be.

tests/e2e/profile_test/get_profile_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	recieved_profile := profile_models.Profile{}
+	response, _ := unauthenticated_client.New().Get(endpoint_address).ReceiveSuccess(&recieved_profile)
+
+	if response.StatusCode != 403 {
+		t.Errorf("Unauthenticated attendee able to access endpoint that requires authentication")
+	}
+}
+
+func TestNonExistantGetProfile(t *testing.T) {
```
Timothy-Gonzalez, line 64: Small thing but Existent is spelled with an e, so this should be `TestNonExistentGetProfile`

Timothy-Gonzalez (approved):
Some small nitpicks but otherwise looks good

Nydauron:
PR moved to #498.

PR #498, opened 2022-11-10
Profile tests

asehgal4:
Added tests to check if an unauthorized user makes a get or post request, if a user submits invalid data for post, and if the user tries to get a profile that doesn't exist.


Timothy-Gonzalez:
Not sure why this and #495 exist but I left some notes there if you want to look at that.

Nydauron (approved):

Timothy-Gonzalez (approved):
LGTM!

PR #501, opened 2022-12-19
Moved auth token generation and verification code to `common`

Nydauron:
Moves all token code out of gateway and auth and into common to allow other services to use said functions if necessary.


Timothy-Gonzalez (approved):
LGTM!

Wish git handled mass renames in a nicer way, but it is what it is...

Timothy-Gonzalez (approved):
LGTM!

asehgal220 (approved):

PR #502, opened 2022-12-20
Testing get for filter and user id

asehgal220:


Nydauron:
Wait until #498 gets merged before reviewing this.

Nydauron (changes requested):
Some tests to add:
- `GET /user/filter/`
  - Add multi-filter test for
  - Try to filter as a normal user and an unauthenticated user
- `GET /user/{id}/`
  - Try to get by id as a normal user and an unauthenticated user
  - Add not found test

PR #503, opened 2022-12-20
Added private events

Nydauron:
Added a field `IsPrivate` to the Event model that is only accessible to Staff and above. Anyone else that searches for an event will not see this field. Any events that have `IsPrivate: true` will also be hidden from the user.


Timothy-Gonzalez:
Looks like this is blocked by #501, so I'll review this once that can be merged.

Timothy-Gonzalez:
#501 just merged, so this needs to be updated to fix the conflicts. Have fun!

Timothy-Gonzalez (changes requested):
#501 just merged, so this needs to be updated to fix the conflicts. Have fun!

asehgal220 (approved):

asehgal220:
From what I can tell, other than the merge conflicts everything looks fine.

Timothy-Gonzalez:
#520 has merge conflicts with this, other than that it looks good.

One small thing might be switching to use common config for token secret, which will be added in #518. We can discuss this also, but I thing is makes sense to have it in common config since it is going to end up being the same `TOKEN_SECRET` in every single config otherwise.

services/event/config/config.go
```diff
@@ -62,5 +64,11 @@ func Initialize() error {
 
 	EVENT_CHECKIN_TIME_RESTRICTED = (checkin_time_res_str == "true")
 
+	TOKEN_SECRET, err = cfg_loader.Get("TOKEN_SECRET")
+
+	if err != nil {
+		return err
+	}
+
```
Timothy-Gonzalez, line 72: Since `TOKEN_SECRET` exists in common now, we shouldn't need to define it like this for every file, and instead just use common_config.

services/event/service/token_service.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/HackIllinois/api/common/authtoken"
+	"github.com/HackIllinois/api/services/event/config"
+)
+
+func IsRequestFromStaffOrHigher(r *http.Request) bool {
+	token := r.Header.Get("Authorization")
+	is_at_least_staff, err := authtoken.IsAuthorized(config.TOKEN_SECRET, token, []authtoken.Role{authtoken.StaffRole, authtoken.AdminRole})
+
+	return err == nil && is_at_least_staff
+}
```
Timothy-Gonzalez, line 15: This seems like a method that might want to be common, which would also make using the token secret much easier.

I don't see anything really event specific here.

services/event/tests/event_test.go
```diff
@@ -878,6 +1041,4 @@ func TestRemoveEventFavorite(t *testing.T) {
 	if !reflect.DeepEqual(event_favorites, &expected_event_favorites) {
 		t.Errorf("Wrong tracker info. Expected %v, got %v", &expected_event_favorites, event_favorites)
 	}
-
-	CleanupTestDB(t)
 }
```
Timothy-Gonzalez, line 1044: Small thing, might be a bigger project to do in a different PR (if so, create an issue for this), it is at the same scale as reformatting everything, but if we could convert all these magic "testid1" to constant variables that would be amazing.

tests/e2e/events_test/get_filtered_events_test.go
```diff
@@ -552,14 +658,88 @@ func TestGetFilteredEventByPoints(t *testing.T) {
...
+	}
+	if response.StatusCode != http.StatusOK {
+		t.Fatalf("Request returned HTTP error %d", response.StatusCode)
+		return
+	}
+
+	expected_events := models.EventList[models.EventDB]{
+		Events: []models.EventDB{
+			{
+				EventPublic: models.EventPublic{
```
Timothy-Gonzalez, line 706: I feel like this testevent1 creation can be extracted out into a function instead of being duplicated 50 times.

Timothy-Gonzalez (changes requested):

services/event/config/config.go
Nydauron: Nice catch. Yeah i set those up prior to making the decision to put `TOKEN_SECRET` in `common`.

services/event/tests/event_test.go
Nydauron: Yeah, this is something that should be resolved in a separate PR. I can go ahead and make an issue for that.

tests/e2e/events_test/get_filtered_events_test.go
Nydauron: This can be also addressed in the issue that I'd say I'd make.

Timothy-Gonzalez (approved):

asehgal220 (approved):
Great Job!

PR #519, opened 2022-12-29
Updating qr endpoints to accept jwt

asehgal220:


Nydauron (commented):
Made some of the following changes:

- Removed references to `auth_config`:
  - `auth_config.Initialize()` does not run on the User microservice (it only runs `user_config.Initialize()`), so `TOKEN_SECRET` should be added to the user config package.
- Added missing error checks
  - Be sure to *always* check for errors first before you use the returned value. Not doing so can result in undefined behavior.

Please check the changes I've added @asehgal220.

Timothy-Gonzalez:
Ok, so in regards to `TOKEN_SECRET`, my pr #518 will move it into common config so that it's much less of a hassle. Probably better than having three separate configs with a `TOKEN_SECRET` in them.

Otherwise, everything else looks good. I'll do a more thorough review once we fix up the token stuff.

Timothy-Gonzalez:
Need to add docs

Nydauron (approved):

Timothy-Gonzalez (approved):

PR #520, opened 2022-12-29
Add caching to CI build job

Nydauron:
Adds caching to CI job to allow for faster runtime.


Timothy-Gonzalez:
This looks good so far, but we still need to cache go dependencies.

Timothy-Gonzalez:
The fact that some tests cache now is an extremely good sign, that's almost a minute shaved off. It also means adding unit tests will be much much much faster.

Timothy-Gonzalez:
Okay, so the goinstall action has issues with caching actual dependencies, and one of the recommended solutions is to do it manually instead. I'm going to try that.

Timothy-Gonzalez:
Under a minute with no changes! That's pretty good. Mostly is just mongo starting now, which I'm guessing can't be cached. Still looks very promising.

Timothy-Gonzalez (approved):

Nydauron:
> Okay, so the goinstall action has issues with caching actual dependencies, and one of the recommended solutions is to do it manually instead.
Huh weird that an `actions/cache` is required even though it literally says on the `actions/setup-go` README that it uses `actions/cache` under-the-hood.

Either way looks great!

asehgal220 (approved):

PR #525, opened 2023-01-24
Add E2E tests for user qr to event checkin connection

Timothy-Gonzalez:
This test encompasses both` GET /user/qr/` and `POST /event/staff/checkin/`, to verify that they always work together.


Nydauron (commented):
Thanks for getting this out quickly! The test looks good.

Could you add two things:
- Add a test that tests if the token fails after 20 seconds (this test should be skippable on CI: e.g. https://stackoverflow.com/questions/24030059/skip-some-tests-with-go-test)
- Add any stuff to the CI to make the test skip (e.g. a non-zero length `CI` env variable such that the integration test knows it's in CI)

Timothy-Gonzalez:
Decided against having an env variable that would never be used anyways (humans are lazy).

Instead, we now have three tests:
1. `GetUserQR` which verifies a token does, and will, expire in a reasonable time.
2. Pre-existing `TestStaffCheckinBadUserTokenExpiredToken` which verifies a expired token cannot be used to check in (and there is also a test for invalid params, malformed token, so every other fail case should be covered)
3. `TestStaffCheckinFromGetUserQR` which verifies that the full workflow works, by fetching a qr and then using it to checkin

Together, this validates that a generated token will expire (1), an invalid token will not work (2), and that a valid token is generated normally and will work to checkin (3).

This may be more complicated than waiting 20 seconds, but we'd never want to wait that long anyways, and a test that never gets run is as good as a test that doesn't exist.

Anyways, should be good now.

Nydauron (dismissed):

asehgal4 (dismissed):

PR #528, opened 2023-02-02
Live leaderboard

Nydauron:
This is mainly for the Siebel dashboard, but this is a new public endpoint: `WS /profile/live/leaderboard/`

Once the connection has been established, the API will send back an initial leaderboard. From then on, the API will send back updated leaderboards whenever it detects updates to player point totals (primarily from `POST /profile/points/award/`). The amount of leaderboards the API can send back due to point value updates is 1 response/second.

The API will also send back an updated leaderboard every 5 minutes of inactivity.

### DIFF
--- a/documentation/docs/reference/services/Profile.md
+++ b/documentation/docs/reference/services/Profile.md
@@ -203,6 +203,47 @@ Request requires no body.
 }
 ```
 
+WS /profile/live/leaderboard/?limit=
+-------------------------
+
+!!! note
+    This is a public endpoint
+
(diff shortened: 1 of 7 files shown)


Timothy-Gonzalez:
Haven't looked over the code yet - but do we really need to be within 1 second live? We could probably get away with once every 10 seconds or every 5 minutes, whichever is more often. It wouldn't really matter other than just not using resources we don't need too.

asehgal220 (approved):

PR #529, opened 2023-02-02
Fix error 2 spew

Timothy-Gonzalez:


Nydauron (approved):
Ig me approving this puts the blame on me. :P

If `flag` ever gets fixed, hopefully, we can revert the exit code check

asehgal220 (approved):

PR #530, opened 2023-02-03
2023 Profile refactor

Nydauron:
- Adds `foodWave` field from `Profile`
- Removes `timezone` field from `Profile`
- Fixed a bug where any user could set their points on profile creation (`POST /profile/`)


Timothy-Gonzalez (approved):
LGTM!

We're removing timezones because we're in person now, right?

asehgal220 (approved):
Looks great!

Nydauron:
> We're removing timezones because we're in person now, right?

Yeah. This was primarily meant for virtual, but since we are not hybrid, this field is not necessary.

PR #534, opened 2023-02-06
Added rsvp data into response from `POST /event/staff/checkin/`

Nydauron:
I never realized that this was a thing until I realized that `POST /checkin/` did this, and since mobile needs registration data, this allows them to access it without having to request via `GET /registration/USERID/` (which would require to decode the JWT user token)

### DIFF
--- a/documentation/docs/reference/services/Event.md
+++ b/documentation/docs/reference/services/Event.md
@@ -487,7 +487,7 @@ Valid values for `status` are `Success`, `InvalidEventId`, `BadUserToken`, `Alre
 {
     "newPoints": 10,
     "totalPoints": 10,
-    "status": "Success"
+    "status": "Success",
     "rsvpData": {
         "id": "github0123456",
         "isAttending": true,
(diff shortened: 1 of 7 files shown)


Timothy-Gonzalez (dismissed):
LGTM!

Do we have a proper type for rsvp data, or is just variable and that's why we use a generic map?

Nydauron:
> Do we have a proper type for rsvp data, or is just variable and that's why we use a generic map?

It's defined within the config itself similar to the registration data. RSVP data typically will like have the user id, whether the user is attending or not, and the associated registration data of that user.

Timothy-Gonzalez:
> > Do we have a proper type for rsvp data, or is just variable and that's why we use a generic map?
> 
> It's defined within the config itself similar to the registration data. RSVP data typically will like have the user id, whether the user is attending or not, and the associated registration data of that user.

I wonder if there's a clean way to build a type from the config... probably not. That's the only nitpick I had, so should be fine.

asehgal220 (approved):
Looks great!

documentation/docs/reference/services/Event.md
```diff
@@ -483,6 +488,11 @@ Valid values for `status` are `Success`, `InvalidEventId`, `BadUserToken`, `Alre
     "newPoints": 10,
     "totalPoints": 10,
     "status": "Success"
```
Nydauron, line 490:
```suggestion
    "status": "Success",
```

PR #535, opened 2023-02-06
Changes `POST /checkin/` and `PUT /checkin/` to take in user token JWT

Nydauron:
Changes expected identification input from `"id"` to `"userToken"`.

Please test this locally to be sure this works


asehgal220 (dismissed):

services/checkin/controller/controller.go
```diff
@@ -58,12 +63,26 @@ func GetCurrentUserCheckin(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	id, err := utils.FetchIdFromSignedUserToken(common_config.TOKEN_SECRET, user_checkin.UserToken)
+	if err != nil {
+		errors.WriteError(w, r, errors.AuthorizationError(err.Error(), "Bad user token."))
+		return
+	}
+
+	user_checkin.ID = id
```
Timothy-Gonzalez, line 84: We can probably remove the id field from here, as it isn't really required.

services/checkin/models/user_checkin.go
```diff
@@ -1,7 +1,8 @@
 package models
 
 type UserCheckin struct {
-	ID              string                 `json:"id"`
+	UserToken       string                 `json:"userToken" bson:"-" validate:"required"`
+	ID              string                 `json:"id" validate:"omitempty"`
```
Timothy-Gonzalez, line 5: Same as above

Timothy-Gonzalez (approved):

Timothy-Gonzalez:
I wasn't able to do local testing, but it looks good. Just a few minor things we can change if we have time, or that should be put into an issue if we don't.

services/checkin/controller/controller.go
Nydauron: While id is not used on input, it is used on output and when returning to the user

common/utils/jwt.go
```diff
@@ -12,6 +12,22 @@ func GenerateSignedToken(secret []byte, data jwt.Claims) (string, error) {
 	return token.SignedString(secret)
 }
 
+func FetchIdFromSignedUserToken(secret string, signed_token string) (string, error) {
```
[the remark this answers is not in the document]
Nydauron: I agree, but the main reason of the duping is to allow for `userId` to be directly casted to a string rather than to a `[]string`. I'll put this into an issue.

While writing this comment out I realize that this line:

https://github.com/HackIllinois/api/blob/f66cf00e69f58a1efa43cfd85db67dfbe44cf3b2/common/utils/jwt.go#L28

does not have any error checking and will cause a panic in case some token does not parse correctly.

services/checkin/models/user_checkin.go
Nydauron: Explanation is in previous comment above

PR #540, opened 2023-02-09
Fixed RSVP confirmtion email to be sent only on `isAttending`

Nydauron:


hackillinoisiostest (approved):
Looks good to me.

Nydauron:
What's also funny is that the approval from @hackillinoisiostest is not a valid approval lmao

PR #542, opened 2023-02-13
Update dev & test config to mirror production

Timothy-Gonzalez:


Timothy-Gonzalez:
Tests need to be updated too

